PR #9, opened 2024-07-31
Switch UI component

sandroguras:


src/lib/components/input/switch.svelte
```diff
@@ -0,0 +1,54 @@
...
+		disabled: isDisabled,
+		defaultChecked: isChecked
+	});
+</script>
+
+<div class="flex items-center">
+	<button
+		use:melt={$root}
+		class="relative outline outline-2 outline-offset-1 outline-gray-700 h-4/5 cursor-default rounded-full bg-zinc-900 transition-colors disabled:outline-offset-0 disabled:bg-gray-700 data-[state=checked]:bg-sky-500 focus-visible:outline-yellow-500 data-[state=checked]:outline-sky-500 data-[state=checked]:outline-offset-0 data-[state=checked]:hover:bg-sky-400 data-[state=checked]:hover:outline-sky-400 data-[state=checked]:focus-visible:outline-yellow-500"
+		id="airplane-mode"
```
deansallinen, line 24: id should be a prop, not hardcoded to 'airplane-mode'

src/lib/components/input/switch.svelte
```diff
@@ -0,0 +1,54 @@
...
+		aria-labelledby="airplane-mode-label"
```
deansallinen, line 25: should probably be derived from the id prop

src/lib/components/input/switch.svelte
```diff
@@ -0,0 +1,54 @@
...
+	} = createSwitch({
+		disabled: isDisabled,
+		defaultChecked: isChecked
+	});
+</script>
+
+<div class="flex items-center">
+	<button
+		use:melt={$root}
+		class="relative outline outline-2 outline-offset-1 outline-gray-700 h-4/5 cursor-default rounded-full bg-zinc-900 transition-colors disabled:outline-offset-0 disabled:bg-gray-700 data-[state=checked]:bg-sky-500 focus-visible:outline-yellow-500 data-[state=checked]:outline-sky-500 data-[state=checked]:outline-offset-0 data-[state=checked]:hover:bg-sky-400 data-[state=checked]:hover:outline-sky-400 data-[state=checked]:focus-visible:outline-yellow-500"
```
deansallinen, line 23: Using outline here provides some artifacting where you see the parent container background through the gap of the outline and the background fill. I wonder if using an inset box shadow e.g. `ring` and increasing the padding of the thumb track is the cleaner approach?

deansallinen, line 23: I'm using `solar-500` for the outline on the buttons, maybe we standardize on one or the other?

src/routes/(dev)/debug/components/+page.svelte
```diff
@@ -127,6 +128,13 @@
 					<Label for="assetInput">Enter token value:</Label>
 					<AssetInput id="assetInput" bind:this={input} bind:value bind:valid bind:min bind:max />
 				</div>
+
+				<div>
+					<h2 class="h2">Switch</h2>
+					<Label for="switchInput">Default:</Label> <Switch />
+					<Label for="switchInput">Disabled:</Label> <Switch isDisabled={true} />
+					<Label for="switchInput">Checked:</Label> <Switch isChecked={true} />
```
deansallinen, line 136: These `<Label for="switchInput>` tags aren't working as labels since the switch doesn't have the id `switchInput` so the label isn't being applied anywhere.

src/routes/(dev)/debug/components/+page.svelte
```diff
@@ -127,6 +128,13 @@
 					<Label for="assetInput">Enter token value:</Label>
 					<AssetInput id="assetInput" bind:this={input} bind:value bind:valid bind:min bind:max />
 				</div>
+
+				<div>
+					<h2 class="h2">Switch</h2>
+					<Label for="switchInput">Default:</Label> <Switch />
```
deansallinen, line 134: Have you thought of how the parent component e.g. a `<form>` will consume the value of the switch?

deansallinen (commented):
Looking good. Few comments for you.

deansallinen:
One more thing: Max mentioned in figma the background on the outlined buttons is transparent, not filled. I'm assuming it's the same for other outlined components like the switches and checkboxes to allow the background to show through.

src/lib/components/input/switch.svelte
sandroguras: makes sense

src/lib/components/input/switch.svelte
sandroguras: makes sense

src/lib/components/input/switch.svelte
sandroguras: I'll try that

src/routes/(dev)/debug/components/+page.svelte
sandroguras: So here's the issue with this approach: 
I initially made the label part of the switch but quickly realized that I was falling into the rabbit hole of making a TW wrapper with a component. Do I expose, say, the GAP property so that the user can widen the distance between the label and the switch? Or should it be justify space-between with a solid GAP, but what about mobile layout?

deansallinen: `<Label>` is a separate component, and layout is handled by the parent container. What I'm saying is once you have the `id` prop on the `<Switch>` component, you can match the `for=` with the `id=` and the Label and Switch will be connected.

sandroguras: Understood, that should be fixed in the next commit

src/routes/(dev)/debug/components/+page.svelte
sandroguras: using a writable store and a custom event. In the parent component i.e. form we will listen for these custom events

src/lib/components/input/switch.svelte
sandroguras: that's done

src/lib/components/input/switch.svelte
sandroguras: that's done

src/lib/components/input/switch.svelte
sandroguras: I tried it, and it works when ring-offset-transparent isn't transparent. When it is - the ring gains one pixel.

sandroguras: that's done

src/routes/(dev)/debug/components/+page.svelte
sandroguras: that's done

src/routes/(dev)/debug/components/+page.svelte
sandroguras: Core functionality established, but need to tie to the form in the main layout

sandroguras:
> One more thing: Max mentioned in figma the background on the outlined buttons is transparent, not filled. I'm assuming it's the same for other outlined components like the switches and checkboxes to allow the background to show through.

That's done as well. However, it messes up the one pixel around the thumb. When the background and the ring-offset are set to transparent, the offset disappears entirely.

src/lib/components/input/switch.svelte
```diff
@@ -0,0 +1,113 @@
+<script lang="ts">
+	import { createSwitch, melt } from '@melt-ui/svelte';
+	import { writable, type Writable } from 'svelte/store';
+	import { onDestroy } from 'svelte';
+
+	interface Props {
```
deansallinen, line 6: `required` should also be an optional prop

src/lib/components/input/switch.svelte
```diff
@@ -0,0 +1,113 @@
...
+		ring-gray-700
+		ring-offset-transparent
+		h-4/5
+		cursor-default
+		rounded-full
+		bg-transparent
+		transition-colors
+		disabled:bg-gray-700
+		disabled:ring-offset-gray-700
+		data-[state=checked]:bg-sky-500
```
deansallinen, line 68: `sky` is the tailwind default not the `skyBlue` that Max has in the designs :)

deansallinen: It's confusing. We should change the name again.

sandroguras:
This branch was merged with the checkbox branch [here](https://github.com/greymass/2nicove/pull/18)

PR #18, opened 2024-08-06
David/components/checkbox

sandroguras:


src/lib/components/input/checkbox.svelte
```diff
@@ -0,0 +1,79 @@
...
+			focus-visible:border-solar-500
+			focus-visible:ring
+			focus-visible:ring-solar-500
+			focus-visible:ring-offset-solar-500
+			"
+			{id}
+			aria-labelledby={ariaLabelledBy}
+		>
+			{#if isChecked}
+				<Check class="size-5" />
```
deansallinen, line 70: How is this class prop being consumed by a component with no exported props? :)

src/lib/components/input/icons/check.svelte
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="2 2 16 16"><path fill="#fff" fill-rule="evenodd" d="M13.8 6.36c-.16 0-.37 0-.6.28l-4.74 5.14-1.75-1.93a.63.63 0 0 0-.6-.2c-.2.06-.59.42-.64.64-.06.22.05.5.2.66L8 13.45c.22.25.73.25.96 0l5.38-5.88c.17-.18.26-.5.17-.73a.93.93 0 0 0-.71-.48Zm0 0"/></svg>
```
deansallinen, line 1: Missing a script tag with props to handle the class. Might also want to allow for other valid svg props. Could do something like:

```
<script lang="ts">
	import type { SVGAttributes } from 'svelte/elements';

	let { class: className }: SVGAttributes<SVGElement> = $props();
</script>
```

deansallinen: Moot though, since we'll be using Lucide as our icon pack throughout the app we'll just install the `lucide-svelte` package and `import {Check} from 'lucide-svelte'` and that'll handle props for us.

src/lib/components/input/checkbox.svelte
```diff
@@ -0,0 +1,79 @@
...
+			{/if}
+			<input use:melt={$input} />
+		</button>
+		<label class="pl-4 font-medium text-magnum-900" for={id}>
```
deansallinen, line 74: `magnum` isn't from our theme

src/lib/components/input/checkbox.svelte
```diff
@@ -0,0 +1,79 @@
...
+			size-5
+			appearance-none
+			border-solid
+			border
+			border-gray-700
+			items-center
+			justify-center
+			rounded
+			bg-transparent
+			text-magnum-600
```
deansallinen, line 50: `magnum` not from our theme

src/lib/components/input/checkbox.svelte
```diff
@@ -0,0 +1,79 @@
...
+			disabled:bg-gray-700
+			data-[state=checked]:bg-sky-500
```
deansallinen, line 52: `sky` should be the value from our theme `skyBlue`

src/lib/components/input/checkbox.svelte
```diff
@@ -0,0 +1,79 @@
...
+			data-[state=checked]:border-sky-500
+			data-[state=checked]:hover:bg-sky-400
+			data-[state=checked]:hover:border-sky-400
+			data-[state=checked]:disabled:bg-gray-700
+			data-[state=checked]:disabled:border-gray-700
```
deansallinen, line 57: Did you try using a gray from our theme? Might be better to use `shark-700` or `mineShaft-700` here

src/lib/components/input/checkbox.svelte
```diff
@@ -0,0 +1,79 @@
...
+		isChecked?: boolean;
+		name: string;
+		id: string;
+	}
+
+	let {
+		isDisabled = false,
+		isChecked = $bindable(false),
+		name,
+		id = 'default-switch-id'
```
deansallinen, line 16: This is a checkbox, not a switch.

deansallinen: Also, I'm wondering if the id should be required with no default. Can't have a dozen `id='default-checkbox-id'` on the page, and we are allowing that to happen with this pattern.

src/lib/components/input/checkbox.svelte
deansallinen: Moot though, the `<label>` shouldn't be included in the checkbox component. We have a separate `Label` component and the parent will control it.

src/lib/components/input/checkbox.svelte
sandroguras: isChecked is an exported prop

src/lib/components/input/checkbox.svelte
```diff
@@ -0,0 +1,79 @@
+<script lang="ts">
+	import { createCheckbox, melt, createSync } from '@melt-ui/svelte';
+	import Check from '$lib/components/input/icons/check.svelte';
+
+	interface Props {
+		isDisabled?: boolean;
+		isChecked?: boolean;
```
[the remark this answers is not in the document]
sandroguras: tweaked

src/lib/components/input/checkbox.svelte
```diff
@@ -0,0 +1,79 @@
...
+		name: string;
+		id: string;
+	}
```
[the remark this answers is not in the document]
sandroguras: tweaked

src/lib/components/input/checkbox.svelte
```diff
@@ -0,0 +1,79 @@
...
+
+	const sync = createSync(states);
+	$effect(() => {
+		sync.checked(isChecked, (v) => (isChecked = v));
+	});
+
+	const ariaLabelledBy = `${id}-label`;
+</script>
+
+<form>
```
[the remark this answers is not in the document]
sandroguras: tweaked

src/lib/components/input/checkbox.svelte
sandroguras: tweaked

src/lib/components/input/checkbox.svelte
sandroguras: when I started there were no colors in the theme yet

src/lib/components/input/checkbox.svelte
deansallinen: On `<Check />` you have `class='size-5'` but `Check` doesn't accept a class prop...

src/lib/components/input/checkbox.svelte
sandroguras: tweaked

src/lib/components/input/checkbox.svelte
sandroguras: tweaked

src/lib/components/input/checkbox.svelte
sandroguras: tweaked

src/lib/components/input/checkbox.svelte
```diff
@@ -0,0 +1,76 @@
...
+<div class="flex items-center">
+	<button
+		use:melt={$root}
+		class="
+			flex
+			size-5
+			appearance-none
+			border-solid
+			border
+			border-gray-700
```
deansallinen, line 45: shouldn't this be `mineShaft`? I don't think we should mix grays

src/lib/components/input/checkbox.svelte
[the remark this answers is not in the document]
deansallinen: not seeing the `required` prop still... also, I think we should change the naming convention here to align better with HTML attributes (e.g. `isChecked` => `checked`). That way we can have `Props` extend `HTMLInputElement` and pass the `...rest` props to the `button` for additional flexibility. What do you think?

src/lib/components/input/checkbox.svelte
```diff
@@ -0,0 +1,76 @@
...
+
+	interface Props {
+		isDisabled: boolean;
+		isChecked: boolean | 'indeterminate';
+		name: string;
+		id: string;
+	}
+
+	let {
+		isDisabled = false,
```
deansallinen, line 14: This should be bindable

src/lib/components/input/checkbox.svelte
```diff
@@ -0,0 +1,76 @@
...
+	>
+		{#if isChecked}
+			<Check />
+		{/if}
+		<input use:melt={$input} />
+	</button>
+	<Label for={id}>
+		Accept terms and conditions.
+	</Label>
+	<input use:melt={$input} {name} />
```
deansallinen, line 75: Why are there two `inputs`?

src/lib/components/input/switch.svelte
```diff
@@ -0,0 +1,88 @@
+<script lang="ts">
+	import { createSwitch, melt, createSync } from '@melt-ui/svelte';
+
+	interface Props {
+		isDisabled: boolean;
+		isChecked: boolean;
+		name: string;
+		id: string;
+	}
```
deansallinen, line 9: Same as checkbox. I think we change the naming convention and extend the standard HTMLInputElement props.

src/lib/components/input/switch.svelte
```diff
@@ -0,0 +1,88 @@
...
+
+	let {
+		isDisabled = false,
```
deansallinen, line 12: Should be bindable

src/lib/components/input/checkbox.svelte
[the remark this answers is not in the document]
sandroguras: I don't get the required prop then? I thought you wanted to make the props not optional and that what I did. And there's no required prop in switch props?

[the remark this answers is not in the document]
deansallinen: https://melt-ui.com/docs/builders/switch

Under Props there is a `required` prop. This should be an optional prop in our component, defaulting to false.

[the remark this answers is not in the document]
sandroguras: Ok, got it. This should be fixed in the next commit

src/lib/components/input/checkbox.svelte
sandroguras: Will be tweaked in the next commit.

src/lib/components/input/checkbox.svelte
sandroguras: Will be fixed for both checkbox and switch in the next commit

PR #19, opened 2024-08-06
style: switch background

deansallinen:
Fixes the background artifacts of the switch and its outline


sandroguras:
can you pr into checkbox branch instead, tnx

PR #29, opened 2024-08-15
Signup & Metamask pages

dafuga:


src/lib/metamask/metamask.ts
```diff
@@ -0,0 +1,120 @@
+import type { EIP6963AnnounceProviderEvent, MetaMaskInpageProvider } from '@metamask/providers';
+
+/**
+ * Check if the current provider supports snaps by calling `wallet_getSnaps`.
+ *
+ * @param provider - The provider to use to check for snaps support. Defaults to
+ * `window.ethereum`.
+ * @returns True if the provider supports snaps, false otherwise.
+ */
+export async function hasSnapsSupport(provider: MetaMaskInpageProvider = window.ethereum) {
```
dafuga, line 10: I wonder if we should move these utility methods to the @wharfkit/wallet-plugin-metamask package 🤔 Either that or maybe we create a new @greymass/metamask-helpers package?

src/routes/[[network]]/(account)/metamask/+page.svelte
```diff
@@ -0,0 +1,92 @@
...
+			<Button onclick={createAccount} disabled={!$isMetaMaskReady || !$installedSnap}>
+				Create Account
+			</Button>
+		</Card>
+	</Grid>
+
+	<Box class="mt-8 rounded-lg bg-gray-100 p-4">
+		<h3 class="mb-2 text-lg font-semibold">How This Works</h3>
+		<p>
+			This signup method allows you to use your MetaMask wallet to interact with Antelope
```
Comment (dafuga) [inline src/routes/[[network]]/(account)/metamask/+page.svelte:86]: Copilot wrote that 😅 I left it as a placeholder for now.

src/lib/state/client/wharf.svelte.ts
```diff
@@ -2,9 +2,12 @@ import { browser } from '$app/environment';
 import { PUBLIC_LOCAL_SIGNER } from '$env/static/public';
 
 import { getContext, setContext } from 'svelte';
-import type { Checksum256 } from '@wharfkit/antelope';
+import type {} from '@wharfkit/antelope';
```
aaroncox, line 5: Should be removed? Not sure if this is even valid hah.

src/lib/state/metamask.svelte.ts
```diff
@@ -0,0 +1,24 @@
+import type { MetaMaskInpageProvider } from '@metamask/providers';
+import { derived, writable, type Readable, type Writable } from 'svelte/store';
+import { isLocalSnap, type Snap } from '../metamask-snap';
+
+export const snapProvider: Writable<MetaMaskInpageProvider | null> = writable(null);
```
aaroncox, line 5: All of these should use Svelte 5 runes - not stores. I know stores are backwards compatible, but I think there's some limitations. Might as well just use runes!

src/lib/state/network.svelte.ts
```diff
@@ -17,13 +17,19 @@ import {
 
 import { calculateValue } from './client/account.svelte';
 
+const snapOriginMap = new Map([
```
aaroncox, line 20: This array probably belongs in the metamask state file or something. I know the origin is being derived and set based on the network, but these values themselves aren't network values.

aaroncox: Actually on second thought - this should be something exported by the plugin, yeah? That way we can inform every app where the origins are, if they wanted to install.

src/routes/[network]/(account)/signup/+page.svelte
```diff
@@ -0,0 +1,28 @@
...
+
+<div class="container mx-auto p-4">
+	<h1 class="mb-4 text-2xl font-bold">Signup</h1>
+	<p class="mb-4">Signup Methods:</p>
+
+	<div class="grid grid-cols-1 gap-4 md:grid-cols-2">
+		{#each signupMethods as method}
+			<div class="card rounded-lg border p-4">
+				<h2 class="mb-2 text-xl font-semibold">{method.name}</h2>
+				<Button onclick={() => goto(method.route)}>{method.name}</Button>
```
Comment (aaroncox) [inline src/routes/[network]/(account)/signup/+page.svelte:24]: Just use links.

src/lib/state/client/wharf.svelte.ts
dafuga: apparently it is 😂

src/lib/state/metamask.svelte.ts
dafuga: Sounds good 👍

dafuga: Actually, I might be missing something, but I don't think that I can use runes the way that I am using the stores here 🤔 Are you saying that you would add a `MetamaskState` class (similar to the `NetworkState` one and that you would add it to the context on page layout mount?

aaroncox: Yeah probably a class that houses the state. That pattern has seemed to work the best.

src/lib/metamask-snap.ts
```diff
@@ -0,0 +1,104 @@
+import type { RequestArguments, MetaMaskInpageProvider } from '@metamask/providers';
```
aaroncox, line 1: `MetaMaskInpageProvider` is unused.

src/routes/[network]/(account)/metamask/+page.svelte
```diff
@@ -0,0 +1,94 @@
...
+	<Grid itemWidth="100%">
+		{#if !metaMaskState?.isMetaMaskReady}
+			<Card class="mb-4">
+				<h2 class="mb-2 text-xl font-semibold">Step 1: Install MetaMask</h2>
+				<p class="mb-2">To get started, you need to install MetaMask:</p>
+				<Button href={'https://metamask.io/download/'} blank>Install MetaMask</Button>
+			</Card>
+		{:else if !metaMaskState.installedSnap}
+			<Card class="mb-4">
+				<h2 class="mb-2 text-xl font-semibold">Step 2: Install Antelope Snap</h2>
```
Comment (aaroncox) [inline src/routes/[network]/(account)/metamask/+page.svelte:72]: Install the "EOS Wallet" snap here.

src/routes/[network]/(account)/metamask/+page.svelte
```diff
@@ -0,0 +1,94 @@
...
+				<p class="mb-2">Install the Antelope Snap for MetaMask:</p>
+				<Button onclick={() => requestSnap(metaMaskState)}>Install Antelope Snap</Button>
+			</Card>
+		{:else}
+			<Card class="mb-4">
+				<h2 class="mb-2 text-xl font-semibold">Step 3: Create an Account</h2>
+				<p class="mb-2">Create your Antelope account:</p>
```
Comment (aaroncox) [inline src/routes/[network]/(account)/metamask/+page.svelte:79]: This is an EOS account (or Jungle or Kylin, etc)

src/app.d.ts
```diff
@@ -13,6 +13,20 @@ declare global {
...
+			detected?: MetaMaskInpageProvider[];
+			providers?: MetaMaskInpageProvider[];
+		};
+	}
+
+	// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
+	interface WindowEventMap {
+		'eip6963:requestProvider': EIP6963RequestProviderEvent;
+		'eip6963:announceProvider': EIP6963AnnounceProviderEvent;
+	}
```
aaroncox, line 29: I don't think we need any of this, do we? Everything seems to work fine without it and the only reason we'd need this is if we reached in and interacted with these elements directly from within unicove.

src/lib/components/select/account.svelte
```diff
@@ -11,9 +11,13 @@
 	import { Stack } from '../layout';
 	import Button from '../button/button.svelte';
 
+	import { goto } from '$app/navigation';
+	import { page } from '$app/stores';
+
 	const context = getContext<UnicoveContext>('state');
 
 	let currentSession = $derived(context.wharf.session);
+	const currentNetwork = $page.params.network || 'eos';
```
aaroncox, line 20: You shouldn't need to default to EOS here. The network is now a required page parameter.

src/lib/state/metamask.svelte.ts
```diff
@@ -0,0 +1,16 @@
+import type { MetaMaskInpageProvider } from '@metamask/providers';
+import { isLocalSnap, type Snap } from '../metamask-snap';
+
+export class MetaMaskState {
```
aaroncox, line 4: Would be neat to create a page to monitor this state. Similar to this:

https://github.com/greymass/2nicove/blob/dev/src/routes/%5Bnetwork%5D/(dev)/debug/state/network/%2Bpage.svelte

A debug page we can go visit to see all the various states, and potentially install/uninstall the plugin. Sort of like the companion dapp inside of the snap, except as a part of unicove for troubleshooting.

Honestly we might be able to get rid of the companion dapp inside of that repo if we just reproduce it here.

aaroncox (approved):

src/app.d.ts
dafuga: Yep! This was needed before but it's no longer needed since we moved a lot of functions to the wallet plugin.

src/lib/state/metamask.svelte.ts
dafuga: Hmm yeah, that could be useful, I'm gonna make an issue for it 👍

PR #31, opened 2024-08-16
Send

aaroncox:


src/routes/[network]/(account)/send/+page.svelte
```diff
@@ -1,40 +1,194 @@
...
+					{context.account.balance?.liquid}
+					<Button disabled={!context.account} onclick={preventDefault(max)} type="button"
+						>Fill Max</Button
+					>
+				{:else}
+					0.0000
+				{/if}
+			</p>
+		</fieldset>
+		<fieldset class="grid gap-3" class:hidden={f.current !== 'memo'}>
```
Comment (dafuga) [inline src/routes/[network]/(account)/send/+page.svelte:167]: Since not everyone will want to leave a memo, should we just display this field at the same time as the confirm button? Similar to what we did on Unicove 1.0?

src/routes/[network]/(account)/send/state.svelte.ts
```diff
@@ -0,0 +1,27 @@
+import { Asset, Name, Serializer } from '@wharfkit/antelope';
+
+export class SendState {
```
Comment (dafuga) [inline src/routes/[network]/(account)/send/state.svelte.ts:3]: I like the idea of putting the form state in a different file 👍

src/routes/[network]/(account)/send/+page.svelte
Comment (aaroncox) [reply src/routes/[network]/(account)/send/+page.svelte]: Yeah potentially, good call. I haven't tackled the review screen really yet but could just move it there.

PR #33, opened 2024-08-19
Side navigation

deansallinen:
Adds sidebar navigation component
Fixed on desktop, hidden on mobile
Updates the root layout to accomodate the new navigation pattern
Removes the PageMargin component as the root layout handles that spacing now
Changes font size to match the designs in Figma (base 12pt) so everything is smaller now


src/app.html
```diff
@@ -1,13 +1,15 @@
...
+	<meta charset="utf-8" />
+	<meta name="viewport" content="width=device-width, initial-scale=1" />
+	<meta name="color-scheme" content="dark" />
+	%sveltekit.head%
+</head>
+
+<body data-sveltekit-preload-data="hover">
+	<div style="display: contents">%sveltekit.body%</div>
+</body>
+
```
aaroncox, line 14: Something happened with formatting here. I ran `make format` and it reverted this back to what it was.

src/lib/components/select/account.svelte
```diff
@@ -6,6 +6,7 @@
 
 	import { chainMapper } from '$lib/wharf/chains';
 	import type { UnicoveContext } from '$lib/state/client.svelte';
+	import { chainLogos } from '@wharfkit/session';
```
aaroncox, line 9: I know it works with session, and this is picky, but I've been trying to pull things from their real libraries rather than the export the session kit does. 

So this one would be `@wharfkit/common`.

aaroncox (approved):
Aside from these minor things, and the actual navigation items needing to be updated like we talked in slack... this is looking real good!

src/lib/components/select/account.svelte
deansallinen: Cool! Changed

src/app.html
deansallinen: Ran `make format`

PR #38, opened 2024-08-21
Send input fixes

deansallinen:
Fixes input props and correctly passes HTMLInputAttributes (e.g. `id`) to the root component
Changes Asset component to use `type=number` for the number keyboard on mobile
Adds PageHeader component for consistent cross-page styling
Adds some additional styling and layout to the Send page
Fixes an issue with buttons in a flex context
Adds an animation on the Send page buttons

### DIFF
--- a/src/lib/components/input/asset.svelte
+++ b/src/lib/components/input/asset.svelte
@@ -106,7 +106,14 @@
 		});
 </script>
 
-<TextInput bind:ref bind:value={input} placeholder={zeroValue.quantity} {autofocus} {...props} />
+<TextInput
+	bind:ref
+	bind:value={input}
+	placeholder={zeroValue.quantity}
+	{autofocus}
+	type="number"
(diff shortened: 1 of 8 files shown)


src/lib/components/input/asset.svelte
```diff
@@ -105,7 +106,14 @@
 		});
 </script>
 
-<TextInput bind:ref bind:value={input} placeholder={zeroValue.quantity} {autofocus} />
+<TextInput
+	bind:ref
+	bind:value={input}
+	placeholder={zeroValue.quantity}
+	{autofocus}
+	type="number"
```
aaroncox, line 114: Just for your reference: this actually shouldn't be a number, it is text. 

The reason is that some numbers EOS deals with are so large, they don't fit into an input typed as number.

aaroncox (approved):

PR #39, opened 2024-08-21
Adding initial settings concept

aaroncox:
Adds the `/settings` page, which contains:

- Language switcher
- Enable Advanced mode
- Enable Debug mode

Then altered the `/send` page to only display debugging information if the debug mode setting is true.


dafuga (approved):
Looks good to me 👍

PR #61, opened 2024-08-24
Ram sell and buy pages

dafuga:


src/routes/[network]/(account)/ram/buy/+page.svelte
```diff
@@ -0,0 +1,117 @@
...
+				action: data.network.contracts.system.action('buyram', buyRamState.toJSON())
+			});
+			alert('RAM purchase successful');
+		} catch (error) {
+			console.error(error);
+			alert('RAM purchase failed: ' + (error as { message: string }).message);
+		}
+	}
+
+	let getRamPriceInterval: NodeJS.Timeout;
```
Comment (aaroncox) [inline src/routes/[network]/(account)/ram/buy/+page.svelte:33]: FWIW this I don't think is a valid type here. I know its going to be removed since the global state keeps the RAM price in sync at all times, but this I think should be:

```ts
let getRamPriceInterval: ReturnType<typeof setInterval>
```

bun.lockb
aaroncox, line 1: What happened here? I don't think the lock update should be needed for anything in here 🤔

aaroncox (commented):
Few random comments of things I noticed. The rest of this is going to change so much that I didn't dive deep into how a lot of its going to be done.

Most of the state in the state.svelte files is likely to be removed since all that calculation is done in the global state already, and just needs to be fetched.

src/routes/[network]/(account)/ram/buy/+page.svelte
Comment (dafuga) [reply src/routes/[network]/(account)/ram/buy/+page.svelte:33]: As you mentioned, I'm removing that code in any case 😄

bun.lockb
dafuga: Ehh I just ran `bun install --yarn` 😅

dafuga: I removed the changes manually, but I think that there is something going on because they are getting readded every time I install the dependencies 🤔

dafuga:
@aaroncox I removed the ram price fetching stuff and am now using the global state instead. We still need what is in the `state.svelte.ts` files though because that's what is used to store the page state.

PR #83, opened 2024-08-28
Initial search box

aaroncox:


src/lib/components/input/search.svelte
```diff
@@ -0,0 +1,85 @@
...
+	let {
+		autofocus = false,
+		network,
+		ref = $bindable(),
+		debug = false,
+		...props
+	}: NameInputProps = $props();
+
+	let searchValue: string = $state('');
+	const searchType = $derived.by(() => {
```
dafuga, line 23: Should we move some of this state logic to a `state.svelte.ts` file?

aaroncox: If it gets to be too much more, sure. I just don't think it needs to do anything than this.

PR #85, opened 2024-08-29
RAM forms followup

dafuga:


src/lib/components/input/assetOrUnits.svelte
```diff
@@ -0,0 +1,96 @@
+<script lang="ts">
```
aaroncox, line 1: I'm not sure this component makes any sense. Why wouldn't the form itself just change what input its using?

aaroncox (approved):

src/routes/[network]/(account)/ram/buy/state.svelte.ts
```diff
@@ -40,18 +52,24 @@ export class BuyRAMState {
 	}
 
 	reset() {
-		this.payer = Name.from('');
-		this.receiver = Name.from('');
-		this.bytes = 0;
-		this.valid = false;
+		this.bytes = undefined;
+		this.tokens = Asset.fromUnits(0, this.chain?.systemToken?.symbol || '4,EOS');
```
Comment (aaroncox) [inline src/routes/[network]/(account)/ram/buy/state.svelte.ts:56]: The value of `this.chain` shouldn't be optional and you shouldn't ever need to default to `4,EOS`. 

If you do, we gotta change something somewhere else. I don't want to scatter the `4,EOS` symbol all over the app.

aaroncox (changes requested):

src/lib/components/input/assetOrUnits.svelte
dafuga: I thought we talked about making this component 😅 IMO having a component like this makes sense if we feel like the asset/units pair of inputs is one that we are going to use in a few places. If you don't think so then I agree that we don't need this component.

src/routes/[network]/(account)/ram/buy/+page.svelte
```diff
@@ -1,38 +1,43 @@
...
 
+	let transactionId: Checksum256 | undefined = $state();
+
 	async function handleBuyRAM() {
 		if (!context.wharf || !context.wharf.session) {
 			alert('Not logged in');
 			return;
 		}
 
+		const actionName = buyRamState.format === 'asset' ? 'buyram' : 'buyrambytes';
```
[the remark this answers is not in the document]
Comment (dafuga) [reply src/routes/[network]/(account)/ram/buy/+page.svelte:31]: Maybe? 🤔 I feel like it's a bit overkill, but if you think that it would be valuable to be able to link to a specific method then I'm not against the idea.

src/routes/[network]/(account)/ram/buy/state.svelte.ts
Comment (dafuga) [reply src/routes/[network]/(account)/ram/buy/state.svelte.ts:58]: it looks like systemToken is optional right now 🤔

src/lib/components/input/assetOrUnits.svelte
aaroncox: I mentioned needing a component for integers, but didn't remember mentioning a component that does both an asset and an integer. The only place I can imagine this being used is on the buy/sell pages for RAM 🤔

src/routes/[network]/(account)/ram/+page.svelte
```diff
@@ -2,10 +2,49 @@
...
+	interface RamPriceData {
+		date: string;
+		value: number;
+	}
+
+	let historicalPrices: RamPriceData[] = $state([]);
+
+	onMount(async () => {
+		try {
+			const response = await fetch('https://unicove-eos.greymass.io/api/marketprice/ram');
```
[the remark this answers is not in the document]
Comment (aaroncox) [reply src/routes/[network]/(account)/ram/+page.svelte:20]: The base URL should also be a environmental variable and not hardcoded. 

Maybe a `API_EOS_METRICS` and `JUNGLE4_API_METRICS` to mirror the others.

src/routes/[network]/(account)/ram/buy/+page.svelte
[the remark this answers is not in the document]
Comment (aaroncox) [reply src/routes/[network]/(account)/ram/buy/+page.svelte:31]: Maybe it is, I'm not entirely sure. I imagined this from the start as being multiple pages with maybe some shared components between them. But now that we're faced with the decision I'm not entirely sure which is best. That's why I figured I'd at least bring it up.

src/routes/[network]/(account)/ram/buy/state.svelte.ts
Comment (aaroncox) [reply src/routes/[network]/(account)/ram/buy/state.svelte.ts]: Hrm it is, but we should require it. If its not set something's wrong (despite what the type says).

src/lib/components/input/assetOrUnits.svelte
dafuga: Maybe we misunderstood each other 😅 I'm actually also having a hard time coming up with other use cases for it 🤔 The only I can think of off the top of my head is that it could be used to buy and sell something like drops.

yarn.lock
```diff
@@ -1,6 +1,7 @@
 # THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
 # yarn lockfile v1
-# bun ./bun.lockb --hash: 4D9DB93CA704CBC9-467373d0eedafba2-E90876A9D21AA761-1c70a670ec9d55a3
+# bun ./bun.lockb --hash: 95FCCD1494BF3AFC-743c08abcaec550a-AB68AA50B0372A99-71027063e5353efe
+
```
aaroncox, line 4: Maybe discard this change to avoid a conflict?

src/lib/components/input/asset.svelte
```diff
@@ -1,7 +1,7 @@
 <script lang="ts">
 	import { Asset, Int64 } from '@wharfkit/antelope';
 	import type { ComponentProps } from 'svelte';
-	import TextInput from './textinput.svelte';
```
aaroncox, line 4: I'm not sure if this was named `textinput` for a specific reason. It might have been to avoid conflicts/confusion with other generic `Text` imports. @deansallinen any insights?

src/lib/components/input/assetOrUnits.svelte
aaroncox: I'm not sure it'd be used there either, since drops are generated/destroyed and you don't ever enter the amount of EOS/RAM.

src/lib/components/input/assetOrUnits.svelte
```diff
@@ -0,0 +1,96 @@
...
+	import { Asset } from '@wharfkit/antelope';
+	import { createEventDispatcher, type ComponentProps } from 'svelte';
```
aaroncox, line 3: Is this ever used?

src/lib/components/input/text.svelte
```diff
@@ -1,11 +1,14 @@
...
 	import type { HTMLInputAttributes } from 'svelte/elements';
+	import { createEventDispatcher } from 'svelte';
 
 	interface TextInputProps extends HTMLInputAttributes {
 		ref?: HTMLInputElement;
 	}
 
 	let { ref = $bindable(), value = $bindable(), ...props }: TextInputProps = $props();
+
+	const dispatch = createEventDispatcher();
```
aaroncox, line 11: Is this ever used?

src/lib/components/input/assetOrUnits.svelte
dafuga: But you could decide to generate drops based on either a unit number or EOS amount in which case you could use that component.

src/lib/components/input/assetOrUnits.svelte
dafuga: No 😄 That's why I hadn't marked the PR as ready for review yet, I still need to do some cleanup.

src/lib/components/input/asset.svelte
deansallinen: I think it was a holdover from when it wasn't in the `/input` directory. I don't know if just `<Text>` would make sense when reading the markup since it's an input... but the filename can change - that's fine. I'd also like to move to using an 'index.ts' file in each folder so we keep consistent import names in our pages. e.g. `import {TextInput, Asset} from '$components/input'`

src/lib/components/input/assetOrUnits.svelte
aaroncox: You'd never input an EOS amount for Droplets I don't think.

src/lib/components/input/asset.svelte
aaroncox: I was mostly worried that a `<Text />` element would conflict with something else. I suppose we could just import it as `InputText` or `TextInput` still though.

src/lib/wharf/chains.ts
```diff
@@ -1,6 +1,8 @@
 import type { Asset } from '@wharfkit/antelope';
 import type { ChainIndices } from '@wharfkit/session';
 
+import { PUBLIC_API_EOS_METRICS } from '$env/static/public';
```
aaroncox, line 4: This should be private and named `API_EOS_METRICS`.

We don't want this endpoint to be publicly known.

src/lib/wharf/chains.ts
```diff
@@ -15,9 +17,12 @@ export interface ChainConfig {
 	name: ChainShortName;
 	features: Record<FeatureType, boolean>;
 	symbol: Asset.SymbolType;
+	ramPrices?: {
+		url: string;
```
aaroncox, line 21: This isn't a `ramPrices` attribute. This API endpoint provides a ton of other metrics, which is why the configuration parameter is called `timeseries`. This should be renamed to reflect that, and probably just be a root property called something like `timeseries_api` or something.

src/routes/[network]/(account)/ram/+page.svelte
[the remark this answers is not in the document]
Comment (aaroncox) [reply src/routes/[network]/(account)/ram/+page.svelte:20]: Like I mentioned in the other comment, this needs to be pulled out and placed inside of a `+page.server.ts` file.

src/routes/[network]/(account)/ram/+page.svelte
```diff
@@ -1,11 +1,66 @@
...
+		if (!data.network.config.features.timeseries) {
+			return;
+		}
+		try {
+			const response = await fetch('https://unicove-eos.greymass.io/api/marketprice/ram');
+			const parsedResponse = await response.json();
+			historicalPrices = parsedResponse
+				.map((price: RamPriceData) => ({
+					date: new Date(price.date),
+					value: Asset.from(price.value / 10000, data.network.chain.systemToken?.symbol || '4,EOS')
```
Comment (aaroncox) [inline src/routes/[network]/(account)/ram/+page.svelte:25]: We talked about removing this and I didn't have a solution at the time - but before this call, just throw an error if the system token isn't defined. We do not want to default to EOS.

src/routes/[network]/(account)/ram/buy/bytes/+page.svelte
```diff
@@ -1,62 +1,57 @@
...
 
 		try {
-			await context.wharf.transact({
-				action: data.network.contracts.system.action('buyram', buyRamState.toJSON())
+			const transactionResult = await context.wharf.transact({
+				action: data.network.contracts.system.action('buyrambytes', buyRamState.toJSON())
 			});
-			alert('RAM purchase successful');
+
+			transactionId = transactionResult?.resolved?.transaction.id;
```
Comment (aaroncox) [inline src/routes/[network]/(account)/ram/buy/bytes/+page.svelte:39]: If we don't get a transaction ID, should this proceed? 

Instead of these optional variables, should we throw an error here if the result or resolved transaction isn't set?

src/lib/components/transaction.svelte
```diff
@@ -0,0 +1,17 @@
...
+	import { transactions } from '$lib/wharf/transact.svelte';
+
+	const { transactionId, network } = $props();
+
+	const transaction = $derived(transactions.find((t) => t.transaction?.id === transactionId));
+</script>
+
+<div class="space-y-4 rounded-lg border border-gray-300 p-4">
+	<h2 class="h2">Transaction Complete</h2>
+	<h3 class="h3">{transaction?.status}</h3>
```
aaroncox, line 11: If the transaction isn't found, should it be showing transaction complete?

src/routes/[network]/(account)/ram/buy/bytes/+page.svelte
```diff
@@ -1,62 +1,57 @@
...
 				buyRamState.payer = context.account.name;
 				buyRamState.receiver = context.account.name;
 			}
-			buyRamState.balance = Asset.from(
-				context.account.balance?.liquid?.value || 0,
-				data.network.chain.systemToken?.symbol
-			);
+
+			if (context.account.balance) {
+				buyRamState.balance = context.account.balance?.liquid;
```
Comment (aaroncox) [inline src/routes/[network]/(account)/ram/buy/bytes/+page.svelte:53]: If this value isn't set, won't this break the form?

aaroncox:
The Jungle 4 version just sits saying "loading ram prices" - but that feature doesn't exist on Jungle and will never load.

src/routes/[network]/(account)/ram/buy/bytes/+page.svelte
```diff
@@ -67,32 +62,39 @@
...
+<form on:submit|preventDefault={handleBuyRAM}>
 	<Stack class="gap-3">
-		<Label for="assetInput">Amount (in bytes)</Label>
-		<Input id="assetInput" bind:value={buyRamState.bytes} />
+		<Label for="bytesInput">Amount to buy (Bytes)</Label>
+		<NumberInput
+			id="bytesInput"
+			bind:value={buyRamState.bytes}
+			placeholder="0"
+			disabled={!context.account}
```
Comment (aaroncox) [inline src/routes/[network]/(account)/ram/buy/bytes/+page.svelte:76]: I wouldn't disable this if there's no account. It might be useful for someone on the form, who's not logged in, to see how much a specific amount would cost by typing it in?

src/lib/components/input/asset.svelte
dafuga: Yeah exactly.

src/lib/components/input/assetOrUnits.svelte
dafuga: I can see some users wanting to buy droplets based on an amount of EOS (for the same reason that some people like to buy a specific EOS amount of RAM).

src/lib/wharf/chains.ts
dafuga: Why do we not want it to be known? If I make it private then I won't be able to use the variable here. I guess we could make an `/api/[network]/ramprices` server-side endpoint for this though.

src/routes/[network]/(account)/ram/buy/bytes/+page.svelte
Comment (dafuga) [reply src/routes/[network]/(account)/ram/buy/bytes/+page.svelte:53]: What do you mean by breaking the form? The balance defaults to 0 so you'll get a validation error if you try to input any amount.

src/routes/[network]/(account)/ram/buy/bytes/+page.svelte
Comment (dafuga) [reply src/routes/[network]/(account)/ram/buy/bytes/+page.svelte:76]: Hmm I guess that makes sense.

src/lib/wharf/chains.ts
aaroncox: Because we don't want people to hit the API directly and abuse it. This needs to be a private API, which only the SSR version of Unicove knows how to access, and essentially proxy/cache it.

src/routes/[network]/(account)/ram/buy/bytes/+page.svelte
Comment (aaroncox) [reply src/routes/[network]/(account)/ram/buy/bytes/+page.svelte:53]: Alright cool. I just wanted to make sure if the balance didn't exist, the default balance wouldn't cause the form to break!

src/lib/wharf/chains.ts
aaroncox: And yes, this should be a server-side endpoint, probably something like:

```
/api/[network]/metrics/marketprice/ram
```

`marketprice` is the name of the data series, and `ram` is the measurement being used.

dafuga: People will be able to abuse the `/api/[network]/metrics/marketprice/ram` though 🤔

dafuga: But yeah, adding caching through cloudflare is probably going to be easier so that's already a good reason to do it this way 👍

src/routes/[network]/(account)/ram/buy/bytes/+page.svelte
Comment (dafuga) [reply src/routes/[network]/(account)/ram/buy/bytes/+page.svelte:39]: That's already happening in the `transact` method. I do need the optionals though because that method is returning a `TransactResult` object and that type marks `resolved` as optional.

src/lib/components/input/assetOrUnits.svelte
dafuga: I'm also not sure why we're still talking about this 😂 I was just pointing out that if we wanted to have a droplets page at some point, you could use a component like this one to let the user generate a certain amount of droplets (you're right that the user wouldn't be buying the droplets, but we could present it as a purchase if we have him buy the RAM needed before generating the droplets).

src/routes/[network]/(account)/ram/+page.svelte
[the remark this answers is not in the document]
Comment (dafuga) [reply src/routes/[network]/(account)/ram/+page.svelte:20]: Do we have a Jungle 4 metrics API right now? 🤔

src/lib/components/input/asset.svelte
deansallinen: Yeah `import TextInput from '$lib/components/input/text'` is fine. 

`import {TextInput} from '$lib/components'` is even better

aaroncox (approved):

src/lib/wharf/chains.ts
dafuga: Oh interersting, I hadn't realized that cf protected us against abuse in any way 🤔

PR #86, opened 2024-08-30
Timeseries Querying

aaroncox:


aaroncox:
Incompatibility between the way API calls are made and the Cloudflare environment we're deploying in.

We'll just build an API and query that instead of building it into the app.

PR #89, opened 2024-09-02
add resources overview page and cpu/net rent

ttwishing:
1. resouces overview page
     a. ram、cpu、net state
     c. link to buyram/sellram
2. cpu/net rent overview page
3. powerup/rex/staking rent page


src/routes/[network]/(account)/resources/+page.svelte
```diff
@@ -1,11 +1,165 @@
...
+			size="{netAvailableSize}kb"
+			used="{netUsagePerc}% Quota used"
+			percentage={netUsagePerc}
+		>
+			<div class="flex flex-col">
+				<Button class="text-blue-400" variant="pill" href="/{network}/resources/net">RENT</Button>
+			</div>
+		</ResourceWrapper>
+	</Grid>
+	<TokensTable {tokenList}>
```
Comment (aaroncox) [inline src/routes/[network]/(account)/resources/+page.svelte:140]: The token balances will appear in another place and this element should not appear on this page.

src/routes/[network]/(account)/resources/types.svelte.ts
```diff
@@ -0,0 +1,9 @@
+export enum RentType {
```
Comment (aaroncox) [inline src/routes/[network]/(account)/resources/types.svelte.ts:1]: Each of these should be defined as features:

https://github.com/greymass/2nicove/blob/94470cf78910d85fcb42bd14731bc025fe8c62f2/src/lib/wharf/chains.ts#L20

Then they should be enabled/disabled on a per-chain basis. Some blockchains will not having staking, rex, and/or powerup.

src/routes/[network]/api/network/+server.ts
```diff
@@ -20,14 +20,15 @@ export async function GET({ fetch, params }) {
 
 	const requests: Promise<ResponseType>[] = [
 		network.resources.v1.ram.get_state(),
-		network.resources.v1.rex.get_state()
+		network.resources.v1.rex.get_state(),
```
Comment (aaroncox) [inline src/routes/[network]/api/network/+server.ts:23]: Here we should check to ensure the features are enabled in the chain configuration. 

If the feature is enabled, we'll make the query. If the feature is disabled, we will not make the query.

src/routes/[network]/(account)/resources/components/overview/tokens.svelte
```diff
@@ -0,0 +1,119 @@
+<script lang="ts">
```
Comment (aaroncox) [inline src/routes/[network]/(account)/resources/components/overview/tokens.svelte:1]: This entire component should also be moved out of the resources section. 

Resources = CPU, NET, and RAM.

These are token balances will live in the "Assets" section of the block explorer pages here:

https://github.com/greymass/2nicove/blob/dev/src/routes/%5Bnetwork%5D/(explorer)/account/%5Bname%5D/balances/%2Bpage.svelte

aaroncox (commented):
This is a great start!

I did a review this morning and I think the next step is removing some of the components that don't belong here. That's the focus of most of my comments so far.

src/routes/[network]/(account)/resources/+page.svelte
Comment (ttwishing) [reply src/routes/[network]/(account)/resources/+page.svelte:140]: @aaroncox OK, I saw the Balance content in Figma and GH task for Resources page, so I added it, I will remove this

src/routes/[network]/api/network/+server.ts
Comment (ttwishing) [reply src/routes/[network]/api/network/+server.ts:23]: @aaroncox I will improve the judgment logic of chain features

src/lib/components/circleprogress.svelte
```diff
@@ -0,0 +1,43 @@
...
+	const girth = 2 * Math.PI * 28;
+	const dasharray = $derived(`${$percentageAnimated * girth},${girth}`);
+
+	$effect(() => {
+		percentageAnimated.set(percentage / 100);
+	});
+</script>
+
+<div class="relative h-[80px] w-[80px]">
+	<svg viewBox="0 0 60 60">
```
dafuga, line 27: Should we move this svg to an asset file?

src/routes/[network]/(account)/resources/components/forms/staking.svelte
```diff
@@ -0,0 +1,129 @@
...
+				min={rentState.min || 0}
+				max={rentState.max || 0}
+			/>
+			{#if !assetValidPrecision}
+				<p class="text-red-500">Invalid number, too many decimal places.</p>
+			{/if}
+			{#if !assetValidMinimum && rentState.amountValue.value > 0}
+				<p class="text-red-500">Amount is below the minimum value</p>
+			{/if}
+			{#if !assetValidMaximum}
```
Comment (dafuga) [inline src/routes/[network]/(account)/resources/components/forms/staking.svelte:104]: When users are not logged in, I think we want to let them enter whatever amount without seeing an`Amount exceeds available balance.` error.

Comment (dafuga) [reply src/routes/[network]/(account)/resources/components/forms/staking.svelte:104]: Actually, I think that using `rentState.insufficientBalance` here should work.

src/routes/[network]/(account)/resources/components/state/state.svelte
```diff
@@ -0,0 +1,58 @@
+<script lang="ts">
```
Comment (dafuga) [inline src/routes/[network]/(account)/resources/components/state/state.svelte:1]: Should we put all components under `src/components`?

src/lib/components/circleprogress.svelte
aaroncox: I'd defer to @deansallinen on where to put these.

deansallinen: Can't imagine this svg would be used outside of this progress component 🤔 I'd say we can leave it in and extract it if we find other implementations

src/routes/[network]/(account)/resources/components/state/state.svelte
Comment (deansallinen) [reply src/routes/[network]/(account)/resources/components/state/state.svelte:1]: Hmm. Not sure what the better pattern is here. I can see us needing route specific components, in which case it's pragmatic to keep them local to the route. But we run the risk of duplication and diverging styles... 

Alternatively if we have global components for everything we lose context and add clutter...

Maybe route specific molecular components based on the global atomic primitives in `src/components`
 can live in their route directory, but new atoms, even if used once, go in `src/components`?

Not sure if there's a right answer

src/routes/[network]/(account)/resources/components/forms/staking.svelte
Comment (aaroncox) [reply src/routes/[network]/(account)/resources/components/forms/staking.svelte:104]: Is there anything to see if they could enter a number? I don't know that there is - there's no estimation or anything updating as the user types.

src/routes/[network]/(account)/resources/components/state/state.svelte
Comment (aaroncox) [reply src/routes/[network]/(account)/resources/components/state/state.svelte:1]: I think that makes sense. 

For this specific component, this was just an element ported over from Unicove 1.0, right? I don't even think its part of the new designs 🤔

src/lib/components/circleprogress.svelte
dafuga: In general, I prefer keeping all assets like SVGs in the same place for maintainability, but in this case I think I agree since this component is needs to be able to update the `stroke-dasharray` value overtime.

src/lib/state/network.svelte.ts
```diff
@@ -9,6 +11,8 @@ import { Contract as DelphiOracleContract } from '$lib/wharf/contracts/delphiora
 import { Contract as SystemContract } from '$lib/wharf/contracts/system';
 import { Contract as TokenContract } from '$lib/wharf/contracts/token';
 
+
+
 import {
```
dafuga, line 16:
```suggestion
import { Contract as TokenContract } from '$lib/wharf/contracts/token';

import {
```

dafuga: Sorry, it was bugging me 😂

src/lib/wharf/chains.ts
```diff
@@ -42,7 +43,11 @@ export const chainConfigs: Record<string, ChainConfig> = {
 			lightapi: true,
 			rex: true,
 			robo: true,
-			timeseries: true
+			timeseries: true,
+			buyram: true,
```
dafuga, line 47: Does `buyram` need to be a feature? 🤔 I feel like RAM is a thing on all networks.

src/routes/[network]/(account)/resources/components/forms/staking.svelte
Comment (dafuga) [reply src/routes/[network]/(account)/resources/components/forms/staking.svelte:104]: Hmm you may have a point that there is nothing to see, but if we are going to let them enter an amount when not logged in then it feels weird to tell them that they don't have the balance for it imo.

src/routes/[network]/(account)/resources/components/forms/powerup.svelte
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}
+
+	function handleSuccessBack() {
+		rentState.reset();
+	}
+</script>
+
+{#if rentState.txid}
+	<Stack>
```
Comment (dafuga) [inline src/routes/[network]/(account)/resources/components/forms/powerup.svelte:84]: There is a new Transaction component that you can use here.

src/routes/[network]/(account)/resources/components/forms/state.svelte.ts
```diff
@@ -0,0 +1,125 @@
...
+	public resourceType: ResourceType;
+	public rentType: RentType;
+	public resourceName;
+	public resourceUnit;
+
+	public payer: Name = $state(Name.from(''));
+	public receiver: Name = $state(Name.from(''));
+
+
+	public coreSymbol: Asset.Symbol = $state(Asset.Symbol.from('4,EOS'))
```
Comment (dafuga) [inline src/routes/[network]/(account)/resources/components/forms/state.svelte.ts:17]: To avoid situations where our pages only work with the EOS symbol, I think you can use a different default symbol here. Aaron and I have been using `UNKNOWN` on the send and ram pages.

src/routes/[network]/(account)/resources/components/forms/state.svelte.ts
```diff
@@ -0,0 +1,125 @@
...
+	public max: number | undefined = $derived(this.balance ? this.balance.value : undefined);
+
+
+	public insufficientBalance: boolean = $derived.by(() => {
+		if (this.cost && this.balance)
+			return this.cost.value > this.balance.value
+		return false;
+	});
+
+	constructor(resource: ResourceType, rent: RentType) {
```
Comment (dafuga) [inline src/routes/[network]/(account)/resources/components/forms/state.svelte.ts:51]: I think we'll need to set the chain here so we can use the correct symbol in the Asset values.

src/routes/[network]/(account)/resources/components/forms/state.svelte.ts
```diff
@@ -0,0 +1,125 @@
...
+		this.resourceUnit = getUnit(resource)
+	}
+
+	reset() {
+		this.payer = Name.from('');
+		this.receiver = Name.from('');
+		this.balance = undefined;
+		this.amount = undefined;
+		this.pricePerUnit = undefined;
+		this.amountValue = Asset.from(0, Asset.Symbol.from('4,EOS'));
```
Comment (dafuga) [inline src/routes/[network]/(account)/resources/components/forms/state.svelte.ts:64]: Once the chain is correctly set in the constructor, you'll be able to use `this.chain.systemToken.symbol` here.

src/routes/[network]/(account)/resources/utils.svelte.ts
```diff
@@ -0,0 +1,43 @@
+import { ResourceType } from "./types.svelte";
```
Comment (dafuga) [inline src/routes/[network]/(account)/resources/utils.svelte.ts:1]: Does this need to be a `.svelte.ts` file?

src/routes/[network]/api/network/+server.ts
```diff
@@ -18,16 +18,34 @@ export async function GET({ fetch, params }) {
...
+	}
+	if (network.config.features.powerup) {
+		powerupStateIndex = addRequest(requests, network.resources.v1.powerup.get_state());
+	}
+	if (network.config.features.staking || network.config.features.rentrex || network.config.features.powerup) {
+		sampleUsageIndex = addRequest(requests, network.resources.getSampledUsage());
+	}
 	if (network.contracts.delphioracle) {
-		requests.push(network.contracts.delphioracle.table('datapoints', 'eosusd').get());
+		tokenStateIndex = addRequest(requests, network.contracts.delphioracle.table('datapoints', 'eosusd').get());
```
Comment (dafuga) [inline src/routes/[network]/api/network/+server.ts:41]: `eosusd` is probably specific to EOS so we need a way to map to the correct delphioracle pairs for other chains (if they exist).

dafuga:
Here is the build is failing:
```
error during build:
--
01:48:38.855 | [vite:load-fallback] Could not load /opt/buildhome/repo/src/lib/components/input/textinput.svelte (imported by src/routes/[network]/(account)/resources/components/forms/staking.svelte): ENOENT: no such file or directory, open '/opt/buildhome/repo/src/lib/components/input/textinput.svelte'
01:48:38.855 | Error: Could not load /opt/buildhome/repo/src/lib/components/input/textinput.svelte (imported by src/routes/[network]/(account)/resources/components/forms/staking.svelte): ENOENT: no such file or directory, open '/opt/buildhome/repo/src/lib/components/input/textinput.svelte'

src/routes/[network]/api/network/+server.ts
Comment (aaroncox) [reply src/routes/[network]/api/network/+server.ts:41]: This was from my code, but you're absolutely right that shouldn't be hardcoded and instead should be defined in a config somewhere per-chain (or maybe env variable). 

I'd propose we just open a new issue for this rather than addressing it in this PR. The problem existed before the PR, and doesn't need to be fixed here?

src/routes/[network]/(account)/resources/utils.svelte.ts
Comment (aaroncox) [reply src/routes/[network]/(account)/resources/utils.svelte.ts:1]: No, just like when you're trying to include a `foo.ts` file and you can just use `import from './foo' without the `.ts` at the end.

src/lib/state/network.svelte.ts
aaroncox: I think `make format` needs to be run. I think that'd pick these sorts of things up

src/lib/wharf/chains.ts
aaroncox: It's different on UX and Proton I believe. Might be good to be able to disable it.

Maybe it could be renamed to `bancormarket` or something - since that's actually the key system that differentiates some of these chains.

src/lib/state/network.svelte.ts
aaroncox, line 1: Just as a heads up, once this PR is merged I am going to take some time to go through and adjust how this network state file works. The addition of these new fields works, but doesn't completely cover all of the situations we are going to need to use it for. 

The reason I'm going to just take care of it after the merge is because I'm having a hard time explaining what needs to be changed 😅

src/routes/[network]/(account)/resources/utils.svelte.ts
Comment (dafuga) [reply src/routes/[network]/(account)/resources/utils.svelte.ts:1]: lol I was talking about the utils.svelte.ts file 😅 I should have left the comment at the file level.

dafuga (approved):

PR #95, opened 2024-09-06
fix asset number check and decimals check

ttwishing:
1. fix case1
input: a
result:  Valid Number:  false

2. fix case2
inpurt: 0.0000001
result: Valid Precision:   false

### DIFF
--- a/src/lib/components/input/asset.svelte
+++ b/src/lib/components/input/asset.svelte
@@ -8,6 +8,7 @@
 		min?: number;
 		max?: number;
 		valid?: boolean;
+		validNumber?: boolean;
 		validPrecision?: boolean;
 		validMinimum?: boolean;
 		validMaximum?: boolean;
@@ -21,6 +22,7 @@
 		max = $bindable(),
 		ref = $bindable(),
 		valid = $bindable(false),
+		validNumber = $bindable(false),
 		validPrecision = $bindable(false),
 		validMinimum = $bindable(false),
 		validMaximum = $bindable(false),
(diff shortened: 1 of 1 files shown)


dafuga:
What is the use case for this? 🤔 I feel like `validPrecision` is what is important on an Asset component and it will only be true if the user input is a valid number.

aaroncox:
Yeah I'm not sure what this is fixing either. 

When visiting the current dev branch at: https://v2.unicove.com/en/eos/debug/input/asset

The above test cases are working as intended. Could you let us know what value you're inputting (in the URL above) isn't working as intended?

ttwishing:
@aaroncox @dafuga At https://v2.unicove.com/en/eos/debug/input/asset,  when you input：0.0000001

input (string):   "0.0000001"
input (number):   "1e-7" // Convert to scientific notation
input (decimals): 0
input (min):      1
input (minUnits): 10000
input (max):      100
input (maxUnits): 1000000
token symbol:     4,FOO
precision:        4
formatted:        0.0000
Asset:            0.0000 FOO

Valid Input:       false       
Valid Number:      true
Valid Precision:   true     // The precision judgment is wrong
Valid Minimum:     false
Valid Maximum:     true

dafuga:
@ttwishing Got it, but the `validNumber` prop still feels a bit redundant though. We could fix the error that you mentioned without adding that prop which is why I was wondering why you added it.

aaroncox:
Good catch on the fact that the input of `0.0000001` wasn't validating properly!

The root problem here was that big.js defaults to using scientific notation in its `.toString()` call. To get the actual number representation string, which is what that length validation uses, the `.toFixed()` should just be called instead. 

I think this solves the problem? 

https://github.com/greymass/2nicove/pull/131

ttwishing:
@aaroncox @dafuga 
Yes, your solution is the best.

I added the validNumber parameter because I want to synchronize the number validation to the caller of the component. Is this parameter necessary?
Another question, when I set validMinimum to 0.0001, the initial value is 0 when no input is given, and satisfactionMinimum will be set to true. Should satisfactionMinimum be false when no input is given? As shown in the figure below
image

aaroncox:
When the page loads and the value is blank, I don't think it should immediately give an error. Errors should only occur after input is entered.

ttwishing:
fixed by https://fix-asset.unicove2.pages.dev/en/eos/debug/input/asset

PR #96, opened 2024-09-06
feat: staking

apporc:


src/lib/components/summary/mvfrsavings.svelte
```diff
@@ -0,0 +1,36 @@
...
+
+	const { action, class: className = '', value, ...props }: MvfrsavingsProps = $props();
+</script>
+
+<Card class="gap-5 text-center {className}">
+	<h3 class="h3">Move out from savings</h3>
+	<Switcher threshold="20rem">
+		<Stack class="gap-0">
+			<p class="caption">owner</p>
+			<p class="h3">{action.data.owner}</p>
```
dafuga, line 26: why don't we use an h3 tag here?

src/routes/[network]/(account)/(staking)/staking/+layout.svelte
```diff
@@ -0,0 +1,59 @@
...
+		<PageHeader title="EOS System Functions" />
+		<AccountNavigation
+			options={[
+				{
+					active: false,
+					href: `/${data.network}`,
+					text: 'Account'
+				},
+				{
+					active: false,
```
Comment (dafuga) [inline src/routes/[network]/(account)/(staking)/staking/+layout.svelte:46]: Shouldn't this one be active on the staking pages?

src/routes/[network]/(account)/(staking)/staking/+page.svelte
```diff
@@ -0,0 +1,89 @@
...
+					>Stake</Button
+				>
+				<Button href="/{networkName}/staking/unstake" variant="secondary" class="text-skyBlue-500"
+					>Unstake</Button
+				>
+			</Switcher>
+		</Card>
+
+		<Card title="Unstaking Balances">
+			<table class="table-auto">
```
Comment (dafuga) [inline src/routes/[network]/(account)/(staking)/staking/+page.svelte:51]: Dean added a new Table component that we can use here.

src/routes/[network]/(account)/(staking)/staking/+page.svelte
```diff
@@ -0,0 +1,89 @@
...
+					</tr>
+				</thead>
+				<tbody>
+					{#each unstaking as record}
+						{#if !record.savings}
+							<tr>
+								<td class="p-4">{record.balance}</td>
+								<td class="p-4 text-right"
+									>{record.date
+										? record.date.toLocaleDateString(undefined, {
```
Comment (dafuga) [inline src/routes/[network]/(account)/(staking)/staking/+page.svelte:65]: We should probably add some helper functions that we can all use for date values like these 🤔

src/routes/[network]/(account)/(staking)/staking/stake/+page.svelte
```diff
@@ -0,0 +1,111 @@
...
+				</Stack>
+			</Card>
+
+			<Card class="gap-5" title="About staking">
+				<Stack class="gap-5">
+					<p class="caption">
+						The APY is an estimate, and may fluctuate based on how many and much others are staking.
+						Your 21 day lockup period starts when you unstake your EOS.
+					</p>
+					<p class="caption">You will never get back less EOS.</p>
```
Comment (dafuga) [inline src/routes/[network]/(account)/(staking)/staking/stake/+page.svelte:106]: But the value of your EOS may go down a lot 😂

src/routes/[network]/(account)/(staking)/staking/stake/state.svelte.ts
```diff
@@ -0,0 +1,103 @@
...
+			owner: this.account!.name!,
+			amount: this.assetValue!
+		});
+		const buyrex = this.network!.contracts.system.action('buyrex', {
+			from: this.account!.name!,
+			amount: this.assetValue!
+		});
+
+		try {
+			const result = await this.wharf!.transact({
```
Comment (dafuga) [inline src/routes/[network]/(account)/(staking)/staking/stake/state.svelte.ts:90]: I think that we want to try to never use `!` in a situation like this one as it could lead to weird errors at runtime.

src/routes/[network]/(account)/(staking)/staking/stake/state.svelte.ts
```diff
@@ -0,0 +1,103 @@
...
+
+		this.minValue = Asset.fromUnits(1, this.network!.chain.systemToken!.symbol).value;
+		this.maxValue = this.account ? getStakableBalance(this.network!, this.account).value : 0;
+	}
+
+	setMaxValue() {
+		this.input?.set(this.stakable);
+	}
+
+	async transact() {
```
Comment (dafuga) [inline src/routes/[network]/(account)/(staking)/staking/stake/state.svelte.ts:79]: I'm kind of thorn about this 🤔 On one hand, it does simplify the svelte component, but there is a part of me that thinks that these state files should only be used to store the state and that transact methods should happen somewhere else.

src/routes/[network]/(account)/(staking)/staking/unstake/state.svelte.ts
```diff
@@ -0,0 +1,90 @@
...
+	public assetValid = $state(false);
+	public assetValidPrecision = $state(true);
+	public assetValidMinimum = $state(true);
+	public assetValidMaximum = $state(true);
+
+	public error: string = $state('');
+	public txid: string = $state('');
+
+	public unstakable: Asset = $derived(
+		this.account ? getUnstakableBalance(this.network!, this.account) : defaultQuantity
```
Comment (dafuga) [inline src/routes/[network]/(account)/(staking)/staking/unstake/state.svelte.ts:28]: I wouldn't use the non-null assertion `!` here either.

src/routes/[network]/(account)/(staking)/staking/utils.ts
```diff
@@ -0,0 +1,122 @@
...
+
+		// TODO: add rexfund
+	}
+	return Asset.fromUnits(balance, network.chain.systemToken!.symbol);
+}
+
+export function getStakedBalance(network: NetworkState, account: AccountState): Asset {
+	let staked = Asset.fromUnits(0, network.chain.systemToken!.symbol);
+	if (account && account.loaded) {
+		const rexInfo = account.account!.data.rex_info;
```
Comment (dafuga) [inline src/routes/[network]/(account)/(staking)/staking/utils.ts:30]: Same idea with the `!` operator here.

Comment (dafuga) [reply src/routes/[network]/(account)/(staking)/staking/utils.ts:30]: It's used in quite a few places so I won't list them all 😄

src/routes/[network]/(account)/(staking)/staking/utils.ts
```diff
@@ -0,0 +1,122 @@
...
+				claimable = Asset.fromUnits(matured, network!.chain.systemToken!.symbol);
+			}
+		}
+	}
+
+	return claimable;
+}
+
+export function getWithdrawableBalance(network: NetworkState, account: AccountState): Asset {
+	// TODO: add rex fund
```
Comment (dafuga) [inline src/routes/[network]/(account)/(staking)/staking/utils.ts:67]: Is the plan to do this in a future PR?

src/lib/components/navigation/sidemenu.svelte
```diff
@@ -21,6 +21,7 @@
 
 		const items = [
 			{ href: `/${network}/account`, text: 'Account' },
+			{ href: `/${network}/staking`, text: 'Staking' },
 			{ href: `/${network}/send`, text: 'Send' },
 			{ href: `/${network}/stake`, text: 'Stake' },
```
aaroncox, line 26: The `/stake` line here should be removed if we're renaming it to `/staking` on line 24.

src/routes/[network]/(account)/(staking)/staking/+layout.svelte
Comment (aaroncox) [reply src/routes/[network]/(account)/(staking)/staking/+layout.svelte:46]: This whole navigation element needs to be deleted anyways.

https://greymass.slack.com/archives/C02EN8NRTL4/p1725663951211209

src/routes/[network]/(account)/(staking)/staking/+page.svelte
```diff
@@ -0,0 +1,89 @@
...
+
+	const context = getContext<UnicoveContext>('state');
+	const { data } = $props();
+	const networkName = String(data.network);
+
+	let staked: Asset = $derived(getStakedBalance(data.network, context.account));
+	let unstaking: Array<UnstakingRecord> = $derived(
+		getUnstakingBalances(data.network, context.account)
+	);
+	let apy = $derived(getAPY(data.network));
```
Comment (aaroncox) [inline src/routes/[network]/(account)/(staking)/staking/+page.svelte:21]: Are these values in this files duplicates of the data that exists in the state.svelte.ts file?

Can that state be used here instead of doing the same calculations?

Unsure of the answer, but wanted to ask the question.

src/routes/[network]/(account)/(staking)/staking/+page.svelte
Comment (aaroncox) [reply src/routes/[network]/(account)/(staking)/staking/+page.svelte:65]: Agreed, maybe we should open an issue to track this one. 

It should be a date that can access what language ParaglideJS is in and potentially use that to determine what language the date is in and how its displayed.

src/routes/[network]/(account)/(staking)/staking/stake/state.svelte.ts
Comment (aaroncox) [reply src/routes/[network]/(account)/(staking)/staking/stake/state.svelte.ts:79]: I have been putting the `transact` call(s) in the page itself so far. Wherever the action is that calls it, so it calls it locally. These sort of state files should just contain the state of the page.

aaroncox:
Bug: The withdraw button doesn't seem to ever let me withdraw.
image

apporc:
> Bug: The withdraw button doesn't seem to ever let me withdraw. image

A typo, thanks.

src/routes/[network]/(account)/(staking)/staking/stake/state.svelte.ts
Comment (apporc) [reply src/routes/[network]/(account)/(staking)/staking/stake/state.svelte.ts:79]: I suggest we re-consider this, i was trying to separate the functional part and representation part from one page. The functional part is not necessarily a `state`, states are just part of it, for example if we move the `transact` out to svelete components, then we need to read values from `StakeState` and fire the transaction there but turn back to update the `StakeState.error` and `StateState.message`, it would be weird in my opinion. Currently nowhere in svelte component updates the states in `StateState` except `$effect` syncing `account` and `network`, states are totally managed inside `StakeState`, this is clear.

src/lib/components/summary/mvfrsavings.svelte
apporc: copied from Aaron's example, i think both are ok.

src/routes/[network]/(account)/(staking)/staking/+layout.svelte
Comment (apporc) [reply src/routes/[network]/(account)/(staking)/staking/+layout.svelte:46]: Yeah i think we are set on the ui design, so i didn't spend much effect here.

src/routes/[network]/(account)/(staking)/staking/+page.svelte
Comment (apporc) [reply src/routes/[network]/(account)/(staking)/staking/+page.svelte:65]: So leave it here for now?

src/routes/[network]/(account)/(staking)/staking/unstake/state.svelte.ts
Comment (apporc) [reply src/routes/[network]/(account)/(staking)/staking/unstake/state.svelte.ts:28]: How would you process the compilation warning about it?
I am using `!` somewhere in this PR for `this.network`. Because it's passed in from constructor, which is `date.network` in corresponding svelte component, which can't be undefined. Check here, 
https://github.com/greymass/2nicove/blob/dev/src/routes/%5Bnetwork%5D/%2Blayout.ts
https://kit.svelte.dev/docs/load

src/routes/[network]/(account)/(staking)/staking/stake/state.svelte.ts
Comment (apporc) [reply src/routes/[network]/(account)/(staking)/staking/stake/state.svelte.ts:90]: Agreed, will add a `if` check here.

src/routes/[network]/(account)/(staking)/staking/utils.ts
```diff
@@ -0,0 +1,122 @@
...
+export const defaultQuantity = Asset.fromUnits(0, defaultSymbol);
+
+export function getStakableBalance(network: NetworkState, account: AccountState): Asset {
+	let balance = Int64.from(0);
+	if (account && account.balance) {
+		if (account.balance && account.balance.liquid) {
+			balance.add(account.balance.liquid.units);
+		}
+
+		// TODO: add rexfund
```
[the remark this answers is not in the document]
Comment (apporc) [reply src/routes/[network]/(account)/(staking)/staking/utils.ts:22]: Yeah, discussing in slack channel.

src/routes/[network]/(account)/(staking)/staking/utils.ts
Comment (apporc) [reply src/routes/[network]/(account)/(staking)/staking/utils.ts:30]: Hmm, the `if` check above checked whether it's loaded.

src/routes/[network]/(account)/(staking)/staking/+page.svelte
Comment (apporc) [reply src/routes/[network]/(account)/(staking)/staking/+page.svelte:21]: No, this is staking overview page, there is no state mutating in the page, so no `OverviewState` class.

src/routes/[network]/(account)/(staking)/staking/+page.svelte
Comment (aaroncox) [reply src/routes/[network]/(account)/(staking)/staking/+page.svelte:65]: Yeah I think so.

src/routes/[network]/(account)/(staking)/staking/unstake/state.svelte.ts
Comment (aaroncox) [reply src/routes/[network]/(account)/(staking)/staking/unstake/state.svelte.ts:28]: Like so?

```
this.account && this.network ? getUnstakableBalance(this.network, this.account) : defaultQuantity
```

src/routes/[network]/(account)/(staking)/staking/stake/state.svelte.ts
Comment (aaroncox) [reply src/routes/[network]/(account)/(staking)/staking/stake/state.svelte.ts:79]: The state though is the forms state - and when you go to `transact`, you're extracting that state into a new format (as actions). 

For the send page, the `toJSON()` method on the state returns the data format needed for the action.

https://github.com/greymass/2nicove/blob/0d47c9e638613cfe3442a047266a20fb1bcd2368/src/routes/%5Bnetwork%5D/(account)/(send)/send/%2Bpage.svelte#L58

It's no longer state, it's now Action JSON.

Overall I agree with your goal of separating state out, and think we could use some more abstraction here - but `transact` shouldn't modify the form state I don't think? 

All that being said, I'm fine merging this PR in as-is without further changes to `transact`. In the future we can try to abstract how we perform transactions and then standardize that new approach across all the different pages/states.

src/routes/[network]/(account)/(staking)/staking/+page.svelte
Comment (dafuga) [reply src/routes/[network]/(account)/(staking)/staking/+page.svelte:65]: Yeah, we can add the helper functions in a future PR.

src/routes/[network]/(account)/(staking)/staking/utils.ts
Comment (dafuga) [reply src/routes/[network]/(account)/(staking)/staking/utils.ts:30]: Yeah, technically you're right, but I think it's safer to just add a condition on line 29 rather than use the `!` operator.

PR #97, opened 2024-09-07
Adding account RAM page

dafuga:


src/routes/[network]/(explorer)/account/[name]/ram/+page.svelte
```diff
@@ -1,23 +1,83 @@
...
+	function bytesToKB(bytes: number | Int64 | undefined): string {
+		return ((Number(bytes) || 0) / 1000).toFixed(2);
+	}
+
+	interface RamPrice {
+		eos?: Asset;
+		usd?: Asset;
+	}
+
+	function calculateRamPriceInEOS(
```
Comment (aaroncox) [inline src/routes/[network]/(explorer)/account/[name]/ram/+page.svelte:16]: Would this function work and let us get rid of these two other functions in this file?

https://github.com/greymass/2nicove/blob/0d47c9e638613cfe3442a047266a20fb1bcd2368/src/lib/state/client/account.svelte.ts#L110-L114

I think you might be able to do:

```ts
calculateValue(Asset.fromUnits(bytes, '3,RAM'), ramprice.eos) // EOS Price
calculateValue(Asset.fromUnits(bytes, '3,RAM'), ramprice.usd) // USD Price
```

src/routes/[network]/(explorer)/account/[name]/ram/+page.svelte
```diff
@@ -1,23 +1,83 @@
...
-			<Code>{JSON.stringify(data.account.network.tokenprice, null, 2)}</Code>
-		{/if}
-		<h3 class="h3">RAM State</h3>
-		<Code>{JSON.stringify(data.account.network.ramstate, null, 2)}</Code>
+	<h3 class="mb-4">Account RAM</h3>
+	<div class="grid grid-cols-1 gap-4 md:grid-cols-3">
+		<Card class="rounded-lg bg-gray-600 p-4 shadow">
+			<h4 class="mb-2 text-lg font-semibold">Total:</h4>
+			<ul class="space-y-2">
+				<li>{bytesToKB(data.account.ram?.max)} KB</li>
```
Comment (aaroncox) [inline src/routes/[network]/(explorer)/account/[name]/ram/+page.svelte:42]: I think we may want a component for this to just standardize how we display RAM across the site.

Something like:

```svelte
<RAMBalance bytes={data.account.ram.max} />
```

That way we can create an element and format it in a similar way across the entire site. Some of the RAM pages you've already built (and the one that Kurt's built) might also be able to make use of it in places.

aaroncox (commented):
Should this page be the same page as the account page here?

https://account-ram-page.unicove2.pages.dev/en/eos/ram

I keep circling back to these overview pages and if they should just be the explorer pages.

src/routes/[network]/(explorer)/account/[name]/ram/+page.svelte
Comment (aaroncox) [reply src/routes/[network]/(explorer)/account/[name]/ram/+page.svelte:42]: Re-reading this - maybe it's a `<RAMElement>` or something. 

We probably want to make these wrappers for all different types of values, like Assets, Integers, etc.

Comment (dafuga) [reply src/routes/[network]/(explorer)/account/[name]/ram/+page.svelte:42]: Makes sense! Will take a crack at that and we can iterate it on it 👍

src/routes/[network]/(explorer)/account/[name]/ram/+page.svelte
Comment (dafuga) [reply src/routes/[network]/(explorer)/account/[name]/ram/+page.svelte:16]: Oh cool, didn't know that that function existed!

dafuga:
@aaroncox I made the changes that you suggested 👍

aaroncox (approved):

PR #98, opened 2024-09-09
add balances overview page

ttwishing:


aaroncox:
There is a lot of duplicated code here for balances, doing things that are already being done in other places by the codebase 😅 

An example of this would be that the balance of all tokens for an account is already being accumulated here:

https://github.com/greymass/2nicove/blob/0d47c9e638613cfe3442a047266a20fb1bcd2368/src/lib/state/client/account.svelte.ts#L42-L46

It automatically refreshes already on an interval by calling `/[network]/api/account/[account]`, which is generated here:

https://github.com/greymass/2nicove/blob/0d47c9e638613cfe3442a047266a20fb1bcd2368/src/routes/%5Bnetwork%5D/api/account/%5B%5Bname%5D%5D/%2Bserver.ts#L38-L59


We also currently don't have token prices - and I'm not sure we will for this first version. We do not want to port all of this price ticker code over from Unicove 1 (which it looks like it was?), we'll want to create a new code for that once it exists.

I'm inclined to just close this pull request. If you want to work on the balances displays, my recommendation would be to just work on the rendering of the data we already have, rather than adding additional data.

ttwishing:
OK, I also think this is not the best solution. I just tried to implement it. I will close it first.

PR #134, opened 2024-09-10
Staking changes

apporc:


src/routes/[network]/(account)/(staking)/staking/utils.ts
```diff
@@ -25,47 +25,46 @@ export function getStakableBalance(network: NetworkState, account: AccountState)
...
-		}
+	const sum: Int64 = unstaking
+		.filter((r) => r.claimable)
+		.reduce((acc, r) => acc.adding(r.balance.units), Int64.from(0));
+	if (sum) {
+		claimable.add(sum);
 	}
 
-	return claimable;
+	return Asset.fromUnits(claimable, network.chain.systemToken!.symbol);
```
Comment (dafuga) [inline src/routes/[network]/(account)/(staking)/staking/utils.ts:59]:
```suggestion
	return Asset.fromUnits(claimable, network.chain.systemToken?.symbol || '4,UNKNOWN);
```

Comment (dafuga) [reply src/routes/[network]/(account)/(staking)/staking/utils.ts:59]: I would do something like that everywhere that you're using the `!`.

src/routes/[network]/(account)/(staking)/staking/withdraw/state.svelte.ts
```diff
@@ -67,24 +67,30 @@ export class WithdrawState {
...
-		actions.push(
-			this.network!.contracts.system.action('withdraw', {
-				owner: this.account!.name!,
-				amount: this.withdrawable!
-			})
-		);
-
 		try {
-			const result = await this.wharf!.transact({
+			if (!this.network || !this.account || !this.account.name || !this.wharf) {
```
Comment (dafuga) [inline src/routes/[network]/(account)/(staking)/staking/withdraw/state.svelte.ts:71]: I still think that this doesn't belong in a state file.

src/routes/[network]/(account)/(staking)/staking/utils.ts
Comment (apporc) [reply src/routes/[network]/(account)/(staking)/staking/utils.ts:59]: Just checked that we are using `this.chain.systemToken.symbol` almost everywhere...

src/routes/[network]/(account)/(staking)/staking/withdraw/state.svelte.ts
Comment (apporc) [reply src/routes/[network]/(account)/(staking)/staking/withdraw/state.svelte.ts:71]: I could use another class name like `WithdrawManager`, instead of `xxxState`, lol. Yeah i agree a pure state should not hold more logics, but here i think what we need is not a pure state. I know how you've wrote other pages, like ram and send pages, read all about them before coding this, but i think this approach in staking is more clear and easy. Tell me what you like, `WithdrawManager` or a pure state and dissecting it in another page file?

src/routes/[network]/(account)/(staking)/staking/utils.ts
Comment (dafuga) [reply src/routes/[network]/(account)/(staking)/staking/utils.ts:57]: I see the error, but I'm saying that I would use ? instead of ! for the reasons that I mentioned on Slack.

Comment (dafuga) [reply src/routes/[network]/(account)/(staking)/staking/utils.ts:57]: Oh yeah, that was my bad, it was because I was using an old version of the wharfkit/common package 😅 I have that all cleaned up in my latest PR 👍

src/routes/[network]/(account)/(staking)/staking/withdraw/state.svelte.ts
Comment (dafuga) [reply src/routes/[network]/(account)/(staking)/staking/withdraw/state.svelte.ts:71]: Renaming it would be a step in the right direction, but my other issue with your approach is that it's not consistent with what we've been doing on the other pages.

Comment (dafuga) [reply src/routes/[network]/(account)/(staking)/staking/withdraw/state.svelte.ts:71]: How about you follow the state class format that we have in the other pages, but you can also have another class to manage the action handling? Maybe it can even take a state class instance as a param? The result would be similar to what you have here, but more consistent with the other pages.

Comment (apporc) [reply src/routes/[network]/(account)/(staking)/staking/withdraw/state.svelte.ts:71]: I don't think state and state manager should be separated. They are just `class's members` and `class's method properties`. Why would we make them two classes?
If you think the `withdrawManager` is the right direction, why not using this approach in all other pages?

Comment (dafuga) [reply src/routes/[network]/(account)/(staking)/staking/withdraw/state.svelte.ts:71]: The reason why you would want two separate classes IMO is because you'd want to decouple the page state from the handling of the transaction since a) we may want to reuse those transaction classes elsewhere and b) I think that having a state class for each page (which is only used to store the page state and nothing else) will make it easier for our future selves to figure out what is going on on any given page.

aaroncox (approved):

src/routes/[network]/(account)/(staking)/staking/withdraw/state.svelte.ts
Comment (aaroncox) [reply src/routes/[network]/(account)/(staking)/staking/withdraw/state.svelte.ts:71]: Just to chime in, the only reason we are using a class is because its required if we want to export those Svelte runes from a separate file. If that wasn't a requirement and we could just export them all as individual runes, we probably would and we wouldn't even have any classes like this (e.g. a "manager").

I'm glad we're having this conversation because we do need to settle on a design pattern for building these pages. I'm not sure what approach is best yet... but the it's kinda awesome that each of these sections is like it's own mini-app and we can experiment with them independently! 

I'm going to merge this... but lets keep this conversation going.

PR #137, opened 2024-09-13
Reworks

aaroncox:
1. Moved `calculateValue` out to the utils file.
2. Added `network.supports('featureType')` and implemented it across many pages. Previously these pages were checking whether a feature like staking existed by looking at the property here `network.config.features.staking`. This new method supports this same lookup directly on the network, but shorter as `network.supports('staking')`.
3. Removed `powerupprice` and `stakingprice` from NetworkState. This values are not singular values, since the price is different for CPU and NET. Those calculations are now moved into the components that need them.

### DIFF
--- a/src/lib/state/network.svelte.ts
+++ b/src/lib/state/network.svelte.ts
@@ -149,9 +149,7 @@ export class NetworkState {
 		this.loaded = true;
 	}
 
-	supports = (feature: FeatureType): boolean => {
-		return this.config.features[feature];
-	};
+	supports = (feature: FeatureType): boolean => this.config.features[feature];
 
 	tokenToRex = (token: AssetType) => {
 		if (!this.rexstate) {
--- a/src/lib/state/network.svelte.ts
+++ b/src/lib/state/network.svelte.ts
@@ -20,7 +20,8 @@ import {
 	chainConfigs,
 	chainMapper,
 	type ChainConfig,
-	type DefaultContracts
+	type DefaultContracts,
+	type FeatureType
 } from '$lib/wharf/chains';
 
 import { tokens } from '../../routes/[network]/api/tokens/tokens';
@@ -98,7 +99,7 @@ export class NetworkState {
 			system: new SystemContract({ client: this.client })
 		};
 
-		if (this.config.features.delphioracle) {
+		if (this.supports('delphioracle')) {
 			this.contracts.delphioracle = new DelphiOracleContract({ client: this.client });
 		}
 	}
@@ -147,6 +148,10 @@ export class NetworkState {
 
 		this.loaded = true;
 	}
+
+	supports = (feature: FeatureType): boolean => {
+		return this.config.features[feature];
+	};
 
 	tokenToRex = (token: AssetType) => {
 		if (!this.rexstate) {
--- a/src/lib/state/network.svelte.ts
+++ b/src/lib/state/network.svelte.ts
@@ -64,15 +64,6 @@ export class NetworkState {
 		}
 		return undefined;
 	});
-	public powerupprice: Asset | undefined = $derived.by(() => {
-		if (this.sampledUsage && this.powerupstate && this.chain.systemToken) {
-			return Asset.from(
-				this.powerupstate.cpu.price_per_ms(this.sampledUsage, 1),
-				this.chain.systemToken.symbol
-			);
-		}
-		return undefined;
-	});
 	public tokenprice = $derived.by(() => {
 		return this.tokenstate ? Asset.fromUnits(this.tokenstate.median, '4,USD') : undefined;
 	});
--- a/src/lib/state/network.svelte.ts
+++ b/src/lib/state/network.svelte.ts
@@ -23,8 +23,8 @@ import {
 	type DefaultContracts
 } from '$lib/wharf/chains';
 
-import { calculateValue } from './client/account.svelte';
 import { tokens } from '../../routes/[network]/api/tokens/tokens';
+import { calculateValue } from '$lib/utils';
 
 export class NetworkState {
 	public chain: ChainDefinition;
--- a/src/lib/state/network.svelte.ts
+++ b/src/lib/state/network.svelte.ts
@@ -54,16 +54,6 @@ export class NetworkState {
 		}
 		return undefined;
 	});
-	public stakingprice: Asset | undefined = $derived.by(() => {
-		if (this.sampledUsage && this.chain.systemToken) {
-			const { account } = this.sampledUsage;
-			return Asset.fromUnits(
-				Number(account.cpu_weight) / Number(account.cpu_limit.max),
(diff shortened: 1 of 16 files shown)


src/lib/state/network.svelte.ts
```diff
@@ -53,25 +54,6 @@ export class NetworkState {
 		}
```
ttwishing, line 54: Is rexprice going to be removed?

src/routes/[network]/(account)/resources/+page.svelte
```diff
@@ -39,26 +33,30 @@
...
+						>SELL</Button
+					>
 				</div>
 			{/if}
 		</ResourceWrapper>
 		<ResourceWrapper resourceState={cpuState}>
-			{#if networkConfig.hasREX || networkConfig.hasPowerUp}
+			{#if context.network?.supports('rentrex') || context.network?.supports('powerup')}
 				<Button class="text-blue-400" variant="pill" href="/{network}/resources/cpu">RENT</Button>
-			{:else if networkConfig.hasStaking}
```
Comment (ttwishing) [inline src/routes/[network]/(account)/resources/+page.svelte:52]: In 1.0, it is if-else, and only one of rent/stake will be displayed. Is it changed here to display both?

src/lib/state/network.svelte.ts
```diff
@@ -148,6 +130,8 @@ export class NetworkState {
 		this.loaded = true;
```
ttwishing, line 130: The current parsing of sampledUsage will cause problems in the calculation of staking price.
`sampledUsage.account.cpu_weight.dividing(sampledUsage.account.cpu_limit.max)`

error: sampledUsage.account.cpu_weight.dividing is not a function

Should I have created a new SampleUsage class that inherits Struct?

src/lib/state/network.svelte.ts
aaroncox: It should, I'm not sure it's accurate either. I think we need to make changes to the resources library to make it accurate though before we can fix it. 

Rex isnt used as much anymore though, so it's not as high of a priority.

src/lib/state/network.svelte.ts
aaroncox: Hrm I thought it these values were typed already.

src/routes/[network]/(account)/resources/+page.svelte
Comment (aaroncox) [reply src/routes/[network]/(account)/resources/+page.svelte:52]: It's possible all 3 methods could be used, so having them independently available without the else statement makes sense.

src/routes/[network]/(account)/resources/components/forms/powerup.svelte
```diff
@@ -28,21 +29,32 @@
...
+				context.network.chain.systemToken
+			) {
 				if (resourceType === ResourceType.CPU) {
 					rentState.frac = context.network.powerupstate.cpu.frac_by_ms(
 						context.network.sampledUsage,
 						Number(rentState.amount)
 					);
+					rentState.pricePerUnit = Asset.from(
+						context.network.powerupstate?.cpu.price_per_ms(context.network.sampledUsage, 1),
+						context.network.chain.systemToken?.symbol
```
Comment (dafuga) [inline src/routes/[network]/(account)/resources/components/forms/powerup.svelte:44]: Do we need the `?` before the `.symbol` here?

src/routes/[network]/(account)/resources/components/forms/staking.svelte
```diff
@@ -18,13 +18,38 @@
 	let quantityInput: AssetInput | undefined = $state();
 
 	$effect(() => {
-		if (context.account && context.network) {
+		if (
+			context.account &&
+			context.network &&
+			context.network.sampledUsage &&
+			context.network.chain.systemToken
```
Comment (dafuga) [inline src/routes/[network]/(account)/resources/components/forms/staking.svelte:25]: Could be missing something, but it looks like `context.network.chain.systemToken` isn't optional in which case we wouldn't need this conditional.

src/routes/[network]/(explorer)/account/[name]/resources/+page.svelte
```diff
@@ -10,12 +10,8 @@
 <div class="space-y-4">
 	<h3 class="h3">RAM Prices</h3>
 	<Code>{JSON.stringify(data.network.ramprice, null, 2)}</Code>
-	<h3 class="h3">Rent Via PowerUp Prices</h3>
```
Comment (dafuga) [inline src/routes/[network]/(explorer)/account/[name]/resources/+page.svelte:13]: We don't want to display these anymore?

src/lib/state/network.svelte.ts
```diff
@@ -148,6 +130,8 @@ export class NetworkState {
...
 	}
 
+	supports = (feature: FeatureType): boolean => this.config.features[feature];
```
dafuga, line 133: ❤️

src/routes/[network]/(account)/resources/components/forms/powerup.svelte
Comment (aaroncox) [reply src/routes/[network]/(account)/resources/components/forms/powerup.svelte:44]: Yeah, `symbolToken` is optional.

https://github.com/wharfkit/common/blob/327b940fe02e1222fa55428b86157d53751c524d/src/common/chains.ts#L26

src/routes/[network]/(account)/resources/components/forms/staking.svelte
Comment (aaroncox) [reply src/routes/[network]/(account)/resources/components/forms/staking.svelte:25]: It isn't optional, see the prior comment.

src/routes/[network]/(explorer)/account/[name]/resources/+page.svelte
Comment (aaroncox) [reply src/routes/[network]/(explorer)/account/[name]/resources/+page.svelte:13]: It isn't calculated on the state object anymore, so I removed it. If we want it here it needs to be calculated.

And as far as prices go, I think thin we need them on this page? This is about the resources on the account, not how much it costs to rent resources.

Comment (dafuga) [reply src/routes/[network]/(explorer)/account/[name]/resources/+page.svelte:13]: Yeah, maybe we don't need to have any prices there. I was just wondering because you left the RAM prices.

src/routes/[network]/(account)/resources/components/forms/powerup.svelte
Comment (dafuga) [reply src/routes/[network]/(account)/resources/components/forms/powerup.svelte:44]: Weird.. it's not optional on the version of the common library that I have installed locally 🤔

Comment (dafuga) [reply src/routes/[network]/(account)/resources/components/forms/powerup.svelte:44]: And I'm on version 1.4.0 which is the latest one. Maybe the `d.ts` file that you released with version 1.4.0 isn't the latest one?

Comment (aaroncox) [reply src/routes/[network]/(account)/resources/components/forms/powerup.svelte:44]: The file seems fine in the released build. Try wiping your node_modules and see if that fixes it. 

It is optional in 1.4.0

Comment (dafuga) [reply src/routes/[network]/(account)/resources/components/forms/powerup.svelte:44]: Yeah, nuking the node_modules folder worked 👍 I guess it was showing me an old version of the d.ts file even if the rest of the package files were 1.4.0 files 🤷

dafuga (approved):
lgtm 👍

PR #138, opened 2024-09-13
RAM pages polishing

dafuga:


src/lib/components/select/select.svelte
```diff
@@ -12,8 +12,8 @@
 	import type { ChangeFn } from '@melt-ui/svelte/internal/helpers';
 
 	interface Props {
-		options: OptionWithImage[];
-		selected: OptionWithImage;
+		options: OptionWithImage<any>[];
```
dafuga, line 15: Couldn't find another way to get these props to work with `OptionWithImage<number>`.

deansallinen: I think we need to rename this type. I was like... why is there an option with an image but there's no image, and it's passed a number :laughing:

deansallinen: I think the type you're looking for here is `SelectOption<number>` imported from `@melt-ui/svelte`. The OptionWithImage is just a convenience wrapper I made up. Maybe we need another type wrapper for `NumberOption`?

src/lib/components/chart/rampricehistory.svelte
```diff
@@ -8,14 +8,16 @@
 	import 'chart.js/auto';
 	import { Card, Stack } from '$lib/components/layout';
 	import { Asset } from '@wharfkit/antelope';
-	import Select from '../select/select.svelte';
+	import Select, { type OptionWithImage } from '../select/select.svelte';
```
deansallinen, line 11: Should probably be `SelectOption` unless we're going to display an image in the dropdown?

src/lib/components/chart/rampricehistory.svelte
```diff
@@ -8,14 +8,16 @@
...
 
 	interface Props {
 		data: { date: Date; value: Asset }[];
 		debug?: boolean;
 	}
 
-	let { data, debug = true }: Props = $props();
+	// Seeing a weird bug with svelte where this component is rerendered with data being undefined
+	// This is a workaround to prevent the component from breaking the page
+	let { data = [], debug = true }: Props = $props();
```
aaroncox, line 20: I'm not seeing any issues here with it removed. This must be your local environment? 

I'd rather not leave code like this here 😅

dafuga: This is super weird, I only see the error when going from the RAM overview page to one of the other RAM pages. You're not seeing the same thing?

src/lib/components/chart/rampricehistory.svelte
dafuga: They both work since the `image` field is optional.

dafuga: Just changed it to `SelectOption` though 👍

src/lib/components/select/select.svelte
dafuga: But the whole point of generics is that you shouldn't need to create a new type wrapper for every value type that you want to support 😂 I thought that `OptionWithImage<string | number>[]` would work, but it also gave me a type mismatch error:

```
Type 'OptionWithImage<string | number>' is not assignable to type 'ListboxOption<number>'.
  Types of property 'value' are incompatible.
    Type 'string | number' is not assignable to type 'number'.
      Type 'string' is not assignable to type 'number'.

Comment (aaroncox) [reply src/lib/components/select/select.svelte:15]: It's already set to `unknown` in the definition, you shouldn't need to use `any` here.

Comment (deansallinen) [reply src/lib/components/select/select.svelte:15]: I guess I'm mostly getting hung up on the name of the type. 

The `OptionWithImage<T>` type is supposed to just extend `<SelectOption<T>` with an additional `image` property. I was expecting them to both be used since its more clear this way when the option should have an image or not.

 Either we make `image` a required property and use that type only when an image is required, or we change the name of the `OptionWithImage` to be less specific and it'll be the catch-all for that component both with and without images. `CustomSelectOption<T>` or something.

What do you think?

Comment (aaroncox) [reply src/lib/components/select/select.svelte:15]: Is the problem here that you aren't typing the array of options?

```
	interface RangeType {
		label: string;
		value: number;
	}

	const range: RangeType[] = [
		{ label: '1D', value: 1 },
		{ label: '1W', value: 7 },
		{ label: '1M', value: 30 },
		{ label: '1Y', value: 365 }
	];

	let selectedRange: RangeType = $state(range[1]);
```

Comment (aaroncox) [reply src/lib/components/select/select.svelte:15]: Nope thats wrong.

Comment (aaroncox) [reply src/lib/components/select/select.svelte:15]: Ah I was close (I think). Is this what dean meant for this part?

```ts
	interface RangeType {
		label: string;
		value: number;
	}

	const range: SelectOption<RangeType>[] = [
		{ label: '1D', value: 1 },
		{ label: '1W', value: 7 },
		{ label: '1M', value: 30 },
		{ label: '1Y', value: 365 }
	];

	let selectedRange: SelectOption<RangeType> = $state(range[1]);

```

src/lib/components/chart/rampricehistory.svelte
aaroncox: No I don't see anything weird when navigating between the RAM page without this code.

src/lib/components/select/select.svelte
deansallinen: this range should just be `SelectOption<number>[]`

deansallinen: Ah, I think part of the problem is the interface for the `select.svelte` component only has `options: OptionWithImage[]` so that's probably the genesis of this whole thing... 

That's just bad naming on my part. I think we should rename it, and just use the extended type exclusively and just make sure the generics work correctly.

deansallinen: FYI created https://github.com/greymass/2nicove/pull/139

dafuga: I added the `<any>` generic because I was getting this type error without it:
```
Type 'CustomSelectOption<unknown>' is not assignable to type 'CustomSelectOption<number>'.
  Type 'unknown' is not assignable to type 'number'.

dafuga: I'm also seeing the type error in your branch @deansallinen

dafuga: It's happening because of the `bind:selected` prop here: https://github.com/greymass/2nicove/pull/138/files#diff-e1eb9e3a9b71568ba6630c6864043389565fc4cd6f073067950689200583b493L125

dafuga: Since the prop is binded, the types have to match both ways. Meaning that the `string` that is passed as a prop has to match the `unknown` type in the component (which it does), but the `unknown` type also has to match the `string` type for cases where data is passed by the Select component to the parent (which it doesn't). That's why the only way that I was able to get it working was by changing the `unknown` type to `any`.

src/lib/components/chart/rampricehistory.svelte
dafuga: 😅 weird.. Maybe there is something different about my environment 🤔

dafuga: I just ran `bun clean` and I'm still seeing the errors.

dafuga: Screenshot 2024-09-14 at 12 28 40 PM

dafuga: Ah actually, the above was redundant because of [this](https://github.com/greymass/2nicove/pull/138/files#diff-e1eb9e3a9b71568ba6630c6864043389565fc4cd6f073067950689200583b493R36) line. If you remove it, you should see the error.

dafuga: I removed the redundent code above 👍

aaroncox (approved):

PR #141, opened 2024-09-16
refactor: use index in token select value

deansallinen:
Fixes type errors and refactors the token select to use an index vs the `tokenbalance` object

Note: generics across components is not well supported. Couldn't get it to work well, so the solution is twofold: 
1) either leave the type of the value key in the options array to be implicitly typed (e.g. `options: SelectOption[]` or `options: ExtendedSelectOption[]` instead of `options: SelectOption<number>[]`, see the examples in the debug page)
2) or, if you're passing in an array of objects that don't extend the `SelectOption` interface - create a custom writable store in the new select component which is typed to the object (e.g. the `TokenSelect` component takes `options: TokenBalance[]` and would implement `writable<TokenBalance>()` for the store. Note: in this case there is a step to make the object an index, so the actual store is `writable<TokenSelectOption>()` but other implementations should be simpler.)

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -23,7 +23,10 @@ export default [
 		files: ['**/*.svelte'],
 		languageOptions: {
 			parserOptions: {
-				parser: ts.parser
+				parser: ts.parser,
+				svelteFeatures: {
+					experimentalGenerics: true
+				}
 			}
 		}
 	},
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -23,10 +23,7 @@ export default [
 		files: ['**/*.svelte'],
 		languageOptions: {
 			parserOptions: {
-				parser: ts.parser,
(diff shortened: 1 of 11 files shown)


eslint.config.js
```diff
@@ -23,7 +23,10 @@ export default [
 		files: ['**/*.svelte'],
 		languageOptions: {
 			parserOptions: {
-				parser: ts.parser
+				parser: ts.parser,
+				svelteFeatures: {
+					experimentalGenerics: true
+				}
```
deansallinen, line 29: Should we remove this now since we couldn't get the generics to work as we intended?

dafuga: Sounds good 👍

dafuga (approved):
Thanks for all your work on this 🙏

PR #146, opened 2024-09-20
fix: type errors

deansallinen:


src/lib/components/select/select.svelte
```diff
@@ -54,120 +45,41 @@
...
+	// However, we still need the method in the TokenSelect component since the type internal to the store is different than the type of the selected prop
 	/** Set the value from a parent */
-	export function set(option: CustomSelectOption<T> | null) {
-		if (!option) {
-			_selected = options[0].value;
-		} else {
-			_selected = option.value;
-		}
-	}
+	// export function set(option: ExtendedSelectOption | null) {
```
dafuga, line 58: Should we just delete this?

src/lib/components/select/select.svelte
```diff
@@ -54,120 +45,41 @@
...
+	// Sync the selected option with the passed in selected prop
 	const sync = createSync({ selected });
-	// $effect(() => {
-	// 	sync.selected(_selected, (v) => (_selected = v || options[0]));
-	// });
+	$effect(() => {
+		sync.selected(_selected, (v) => (_selected = v || options[0]));
+	});
 
+	// NOTE: I'm pretty sure this isn't needed here since the selected prop is already being synced using the $effect above.
```
dafuga, line 54: Should we just delete it if it's not needed?

src/lib/components/select/elements/item.svelte
```diff
@@ -0,0 +1,53 @@
...
+	import { Check } from 'lucide-svelte';
+	import type { Readable } from 'svelte/store';
+	import type { ExtendedSelectOption } from '../types';
+
+	interface Props {
+		variant?: 'pill' | 'form';
+		id: string;
+		// Upstream bug: https://github.com/melt-ui/melt-ui/issues/974
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		option: any;
```
dafuga, line 12: Just spent 10 mins trying to get this working in your branch and I'm also having a hard time 😅 I guess this is fine for now 😂

src/lib/components/select/select.svelte
```diff
@@ -41,7 +33,6 @@
 		states: { open, selected, selectedLabel },
 		helpers: { isSelected }
 	} = createSelect({
-		// defaultSelected: _selected || options[0].value,
```
dafuga, line 44: We no longer need this? (just want to make sure that it wasn't removed by mistake 😄)

src/routes/[network]/(dev)/debug/components/sections/select.svelte
```diff
@@ -1,27 +1,31 @@
...
-	const options = [
+	const options: SelectOption[] = [
 		{ value: 30, label: '30d' },
 		{ value: 60, label: '60d' },
 		{ value: 90, label: '90d' },
 		{ value: 365, label: '365d' }
 	];
-	let selected = $state(options[1]);
+	let selected = $state(options[0]);
+	// let selectedOption = $derived(options[selected]);
```
Comment (dafuga) [inline src/routes/[network]/(dev)/debug/components/sections/select.svelte:18]:
```suggestion
```

dafuga (approved):

src/lib/components/select/select.svelte
deansallinen: Yeah, just wanted to make sure there wasn't a use case I was missing since @aaroncox added this earlier.

src/lib/components/select/select.svelte
deansallinen: Good to remove. Only need it in simple implementations, and ours isn't simple ha

src/lib/components/select/elements/item.svelte
deansallinen: Feels dirty, but yeah, the type from the melt-ui library is bugged so not much we can do right now.

src/lib/components/select/select.svelte
deansallinen: Setting the value can be done by just changing the value of `selected` in the parent. e.g. `selected = option[2]` The `_selected` prop in the `select.svelte` component will then update and get synced with the `selected` store of the `melt-ui` component.

Issue #52, opened 2024-08-23
[OP]-Information about tokens this account staked page


sandroguras:
### URL
`/en/[network]/accounts/[account]/staked`

### Purpose
Displays information about tokens this account has staked

### On page
- [x] General staking information/chart
- [ ] Total EOS staked
- [x] Total earned
- [ ] Staking related activity

### Links to
to /[network]/stake page

Issue #53, opened 2024-08-23
[CP]-Management of token staking page


sandroguras:
### URL
`/en/[network]/stake`

### Purpose
Management of token staking

### On page
- [ ] https://github.com/greymass/2nicove/issues/63

### Links to

sandroguras:
https://github.com/greymass/2nicove/pull/96

PR #164, opened 2024-10-09
staking

apporc:
#52


apporc:
The types check failed, but i think it's bug of the checker. There is a `this.network` for every warning it reports.

apporc:
#53

aaroncox (approved):

PR #166, opened 2024-10-10
add acount overview pages

ttwishing:


src/routes/[network]/(account)/accounts/[name]/+page.svelte
```diff
@@ -0,0 +1,107 @@
...
+
+<Stack>
+	<BalanceCard {totalTokenValue} {totalUsdValue} {systemTokenMeta}></BalanceCard>
+
+	{#if tokenBalances.length}
+		<Stack>
+			<h2 class="h2">Tokens</h2>
+			<TokenTable {tokenBalances}>
+				{#snippet transferIntent(balance: TokenBalance)}
+					<a class="text-skyBlue-500" href="/{network}/send?quantity={balance.asset}">Send</a>
```
Comment (deansallinen) [inline src/routes/[network]/(account)/accounts/[name]/+page.svelte:56]: I noticed this renders a space in the query param (e.g. `quantity=0 EOS`) - not sure if that's going to be valid

Comment (ttwishing) [reply src/routes/[network]/(account)/accounts/[name]/+page.svelte]: There shouldn't be a send operation here, so I removed this code first.

src/routes/[network]/(explorer)/account/[name]/+page.svelte
```diff
@@ -1,9 +1,183 @@
...
+					{@render card(data.account.net)}
+				{/if}
+			</Grid>
+		</Stack>
+	{/if}
+{/if}
+
+<br />
+<br />
+<br />
```
Comment (aaroncox) [inline src/routes/[network]/(explorer)/account/[name]/+page.svelte:180]: I'm assuming they shouldn't be here? 

If we need formatting here we could use CSS.

src/routes/[network]/(explorer)/account/[name]/activity/+page.svelte
```diff
@@ -1,34 +1,48 @@
...
+				</td>
+				<td>
+					{action.action}
+				</td>
+				<td>
+					{JSON.stringify(action.data, null, 2)}
+				</td>
+				<td class="text-right">
+					<div class="flex">
+						<Button variant="pill" href="/{data.network}/transaction/{action.id}">detail</Button>
```
Comment (aaroncox) [inline src/routes/[network]/(explorer)/account/[name]/activity/+page.svelte:39]: This button can be removed, and the link can be done on the `action.id` in the first column.

src/routes/[network]/(explorer)/account/[name]/activity/+page.svelte
```diff
@@ -1,34 +1,48 @@
...
+						<Button variant="pill" href="/{data.network}/transaction/{action.id}/{action.seq}"
+							>seq</Button
```
Comment (aaroncox) [inline src/routes/[network]/(explorer)/account/[name]/activity/+page.svelte:41]: This link can be removed without replacing it for now.

src/routes/[network]/(explorer)/account/[name]/balances/+page.svelte
```diff
@@ -1,12 +1,33 @@
...
+							<div class="flex h-6 w-6 items-center justify-center rounded-full bg-black">
+								{#if balance.metadata?.logo}
+									<img class="h-5 w-5" src={balance.metadata?.logo} alt="LOGO" />
+								{/if}
+							</div>
+							<span>{balance.asset.symbol.name}</span>
+						</div>
+					</td>
+					<td>
+						{balance.asset.value.toFixed(balance.asset.symbol.precision)}
```
Comment (aaroncox) [inline src/routes/[network]/(explorer)/account/[name]/balances/+page.svelte:27]: I think `balance.asset.quantity` will do exactly the same as what this is doing. I don't believe we need to use `toFixed` and then pass in the precision?

src/lib/state/client/account.svelte.ts
```diff
@@ -205,9 +204,17 @@ export function getBalance(network: NetworkState, sources: DataSources): Balance
 		}
 	}
 
+	if (ramResources && network.ramprice) {
+		const asset = Asset.from(`${ramResources.max.dividing(1000)} RAM`);
+		const ramValue = calculateValue(asset, network.ramprice.eos);
+		ram.units.add(ramValue.units);
+		total.units.add(ramValue.units);
```
aaroncox, line 211: I don't think we want to add the EOS value of the accounts RAM to their total EOS balance.

This will cause the accounts total EOS balance to go up and down, which may cause concern to users as they think the amount of EOS they own is changing.

aaroncox (commented):
A good step forward to get these pages under control! I left a few comments on things that could be changed or ways we could further improve the code 👍

src/routes/[network]/(explorer)/account/[name]/+page.svelte
```diff
@@ -1,9 +1,183 @@
...
+		<Card>
+			<div class="flex items-center gap-5">
+				<div class="flex h-14 w-14 items-center justify-center rounded-full bg-[#303338]">
+					{#if data.network.tokenmeta && data.network.tokenmeta.length}
+						<img class="h-6 w-6" src={data.network.tokenmeta[0].logo} alt="LOGO" />
+					{/if}
+				</div>
+				<Stack>
+					<p>Total {data.account.balance?.total.symbol.name || ''} Balance</p>
+					<h3 class="h3">{data.account.balance?.total.value || 0}</h3>
```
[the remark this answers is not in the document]
Comment (ttwishing) [reply src/routes/[network]/(explorer)/account/[name]/+page.svelte:47]: OK, i will improve this

src/routes/[network]/(explorer)/account/[name]/+page.svelte
Comment (ttwishing) [reply src/routes/[network]/(explorer)/account/[name]/+page.svelte:180]: This is just test code, I will remove it

src/lib/state/client/account.svelte.ts
ttwishing: 1. First, value includes ram, so should balance also include ram? I am not sure if they are symmetrical
2. When I use resource->cpu/net state token, the value of EOS Staked balance is changing. So when I buy or sell ram, should the value of EOS RAM also change and be displayed?

ttwishing: https://github.com/greymass/2nicove/blob/cca8ad5164b2c32c737893267fb2094dfa44a207/src/lib/state/client/account.svelte.ts#L139
The calculation here is based on eos, should it be changed to usd?

src/routes/[network]/(explorer)/account/[name]/balances/+page.svelte
Comment (ttwishing) [reply src/routes/[network]/(explorer)/account/[name]/balances/+page.svelte:27]: Yes, I will fix this.

src/routes/[network]/(explorer)/account/[name]/activity/+page.svelte
Comment (ttwishing) [reply src/routes/[network]/(explorer)/account/[name]/activity/+page.svelte:41]: If the activity page continues to improve, should it be paginated?

Comment (aaroncox) [reply src/routes/[network]/(explorer)/account/[name]/activity/+page.svelte:41]: These two links were going to different views of a transaction, one of which was experimental. The 1st link goes to the full transaction with all actions, and the 2nd link went to a page only displaying this specific action trace from within the transaction.

For the sake of speeding up a release, we should just focus on the 1st link and omit the 2nd.

src/lib/state/client/account.svelte.ts
aaroncox: 1. I am not convinced they need to be symmetrical. The reason being is the USD/FIAT value of the account will always be changing, as both the price of EOS and RAM fluctuate over time. Users expect this. What users don't expect is to see the EOS balance of their account decrease in value, which would indicate that something is taking tokens from their account. So for that reason I think its OK to include RAM in the USD/FIAT value, but exclude RAM from the EOS value of an account. The EOS value of the account represents the number of EOS tokens you have, not the value of your account in EOS tokens.

aaroncox: > When I use resource->cpu/net state token, the value of EOS Staked balance is changing. So when I buy or sell ram, should the value of EOS RAM also change and be displayed?

That's a good point. Should we remove the USD/FIAT value of the staked balance, on the staking page? 

Perhaps when a user visits this page what they are really looking for is the balance of EOS tokens over time, regardless of its USD/FIAT value.

aaroncox: > The calculation here is based on eos, should it be changed to usd?

No, I think its missing a step.

What needs to happen here is:

- Convert EOS bytes (`ramResource.max`) into the EOS value, which is happening on lines 140 and 141.
- Convert EOS value into USD value, **this step is missing**.
- Add USD value to `ram.units` and `total.units`.

ttwishing: I removed RAM from the balance, and added the eso->usd step to RAM in the value. From the perspective of staking, it is still the user's eos balance, and RAM should be counted in the usd balance. If so, should the third-party tokens also be counted in the usd balance?

src/routes/[network]/(explorer)/account/[name]/activity/+page.svelte
Comment (ttwishing) [reply src/routes/[network]/(explorer)/account/[name]/activity/+page.svelte]: Should the loading of the activity list be changed to paging loading?

src/lib/state/client/account.svelte.ts
aaroncox: Yes, eventually I think the USD value of the third-party tokens should be included.

I don't believe we have data on the value of those tokens currently though, so for the time being - we'll just use the EOS token and RAM.

src/routes/[network]/(explorer)/account/[name]/activity/+page.svelte
Comment (aaroncox) [reply src/routes/[network]/(explorer)/account/[name]/activity/+page.svelte]: Yes, I believe we do need to add pagination to those activity list pages.

PR #168, opened 2024-10-16
Homepage

deansallinen:


aaroncox:
Some notes:

- [ ] Hitting command+k to bring up the search launches it twice.
- [ ] We should probably add a "search" button to the pop, it's unclear you should hit enter or click the icon.
- [ ] It could be nice to have an "X" on recent search results to clear an individual item.
- [ ] A "Clear recent searches..." would be a nice option on the search menu.
- [ ] On the homepage, when scaling down to mobile sizes, the blockchain logo should probably center above the content.

aaroncox (approved):

PR #175, opened 2024-10-19
Implemented RAM Figma designs and other enhancements

dafuga:


dafuga:
> Some notes:
> 
> * [ ]  Overview: The debugging information on the buy/sell pages should only be displayed if the "Enable Debug Mode" in the settings are is enabled.
> * [ ]  Overview: The RAM supply (~53gb) is incorrect, I'm not sure where this value is coming from.
> * [ ]  Overview: The RAM marketcap is incorrect too.
> * [ ]  Overview: The RAM chart seems to have two different backgrounds that are different colors.
> * [ ]  Overview: The RAM chart shouldn't be as tall as it is, the designs in Figma show it much shorter in height.
> * [ ]  Overview: The font sizing, spacing, etc when compared to the Figma designs doesn't match.
> * [ ]  Buy/Sell: The back button + header doesn't match the designs
> * [ ]  Buy/Sell: The formatting of the details window doesn't match designs
> * [ ]  Buy/Sell: The maximum width of these forms needs to be set.

Oh cool! I hadn't realized that new RAM designs had been added recently. Will implement those in the coming days 👍 

And [this](https://github.com/greymass/2nicove/pull/175/files#diff-3acf3ea33d5ee565c4b8d0539f127bcae822ce1d1477edf0f71fed8a6871d133R23-R33) is how I'm calculating the RAM supply and market cap 🤔 Is there some code that I can look at that would show the needed calculation?

aaroncox:
> And [this](https://github.com/greymass/2nicove/pull/175/files#diff-3acf3ea33d5ee565c4b8d0539f127bcae822ce1d1477edf0f71fed8a6871d133R23-R33) is how I'm calculating the RAM supply and market cap 🤔 Is there some code that I can look at that would show the needed calculation?

Yeah, those aren't right. That's the RAM connector values used to determine the current price, and isn't involved in either the supply or market cap.

You can see more accurate values here: https://ramseos.io

The `global` table on the `eosio` contract I believe has the values we need:

image

The market cap would be based on the maximum value, multiplied by the value of RAM. That would result in the market cap in EOS value, which could then be multiplied by the USD price of EOS, to get the USD marketcap value.

dafuga:
@aaroncox until I can figure out where to get that value, I'll just hard code the 409GB RAM supply. When I do that, the market cap seems to be more or less inline with what I'm seeing on the other websites.

dafuga:
> sed on the maximum value, multiplied by the value of RAM. That would result in the market cap in EOS value, which could then be multiplied by the USD price of E

Oh awesome! Will implement that tomorrow 👍

PR #177, opened 2024-10-23
Updates to signup pages

dafuga:


aaroncox:
Broken on mobile somehow.

image

Shows Anchor text, with MetaMask logo - and when I click it loads the MetaMask instructions.

aaroncox:
Other thoughts:

- [ ] Between the recommended wallet at the top, and the list of other wallets, I think some sort of header needs to exist that says "Other [Type] Wallets".
- [ ] The wallet configuration needs to have which chains it supports. If you go to signup on Jungle 4, it's just going to be Anchor and MetaMask pretty much. If you go to sign up on WAX, the Cloud Wallet needs to be included, etc. 
- [ ] The download links to each wallet need to be more precise locations, not just their homepage. Dropping them on the homepage is probably going to create a suboptimal experience, requiring a ton of extra clicks to figure out where to download it. Like for Anchor on mobile it needs to go to the app store page.
- [ ] The wallet links should probably open in a new tab.
- [ ] For each wallet, do we want to describe feature of account creation somehow in the design? Like when we recommend Anchor, before even clicking "Continue with Anchor Mobile", should it display information about how "Anchor allows you to securely create an account using in-app purchases" or something? This is the onboarding flow, so maybe we should focus on what the user should expect before making their selection.

dafuga:
> Broken on mobile somehow.
> 
> image
> Shows Anchor text, with MetaMask logo - and when I click it loads the MetaMask instructions.

I'm super confused by this 😅 if you log the recommendedWallet values you can see that we're passing the Anchor Mobile logo. Not sure why it's displaying the MetaMask one..

dafuga:
Ahhhh I found the issue, it's because the platform detection is happening server side 😅

dafuga:
Still not sure why it's happening, but making it so the page doesn't get rendered until the platform is detected in the browser at least fixes the issue.

PR #178, opened 2024-10-24
Resources page refactor and ui update

ttwishing:


aaroncox:
Some feedback:

- [ ] The Resources navigation item on the left sidebar should be displayed when the user has "Advanced mode" enabled in the settings.
- [ ] We should add the users EOS token balance somewhere on this page. Perhaps we add it to the bottom section between "Usable for" and "Total cost".
- [ ] The debugging information at the bottom of the page should only be shown when the user has "Enable debugging info" from the settings page.

aaroncox (approved):

PR #179, opened 2024-10-24
Fix default asset symbol in overview page

apporc:


src/routes/[network]/(account)/(staking)/staking/stake/manager.svelte.ts
```diff
@@ -51,7 +50,7 @@ export class StakeManager {
 			this.network = network;
 			changed = true;
 		}
-		if (account?.name !== this.account?.name) {
+		if (this.account !== account) {
```
Comment (dafuga) [inline src/routes/[network]/(account)/(staking)/staking/stake/manager.svelte.ts:53]: Can you compare two account states like this? 🤔

Comment (apporc) [reply src/routes/[network]/(account)/(staking)/staking/stake/manager.svelte.ts:53]: I think it's ok, for your reference, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Equality

this.account and account should be the same object, it then is true.

aaroncox (approved):

PR #189, opened 2024-11-05
Remove the AvailableRam component of ram module

ttwishing:
Resources pages also have the same components, can they be unified?


aaroncox:
I think this makes sense, removing the duplicate element. I'll leave the final call up to @dafuga since he was driving here though.

dafuga:
Yeah, this looks good to me 👍 Thanks @ttwishing !

dafuga (approved):

PR #194, opened 2024-11-07
Fixing RAM chart

dafuga:


src/lib/components/chart/rampricehistory.svelte
```diff
@@ -23,17 +23,32 @@
...
 	let selectedRange: ExtendedSelectOption = $state(range[1]);
 
 	let dataRange = $derived.by(() => {
 		if (data.length === 0) return [];
 		const rangeEndDate = dayjs(data[0].date);
 		const rangeStartDate = rangeEndDate.subtract(Number(selectedRange.value), 'day');
-		return data.filter(({ date }) => dayjs(date).isAfter(rangeStartDate));
+		const filteredData = data.filter(({ date }) => dayjs(date).isAfter(rangeStartDate));
+
+		// If we have more points than MAX_NUM_POINTS, sample them evenly
```
deansallinen, line 40: Did you try the chartjs decimation feature?

https://www.chartjs.org/docs/latest/general/performance.html#decimation

src/lib/components/select/select.svelte
```diff
@@ -47,8 +47,12 @@
 
 	// Sync the selected option with the passed in selected prop
 	const sync = createSync({ selected });
+	let lastSelected: ExtendedSelectOption | undefined = $state();
 	$effect(() => {
-		sync.selected(_selected, (v) => (_selected = v || options[0]));
+		if (JSON.stringify(_selected) !== JSON.stringify(lastSelected)) {
```
aaroncox, line 52: Is there a key/index we could use here as opposed to stringify-ing?

dafuga: Since this value could be anything, this is the only comparison method that I can thing of that should always work. Lmk if you have a better idea though.

src/lib/components/chart/rampricehistory.svelte
dafuga: I had no idea that was a thing 😄 Thanks!

dafuga: Ok so the decimation feature isn't going to be as easy to use as I first thought 😅 It doesn't currently work with our chart because we're not defining both the x and y coordinates (we're just defining labels and an array of data). There is probably a way to get it working, but I don't think it's worth the effort at this point (the filtering that I added already seems to work great).

PR #197, opened 2024-11-08
activity action page

ttwishing:


src/lib/utils/transactions.ts
```diff
@@ -0,0 +1,207 @@
...
+		}
+		return super.getActionName(currentAccount, action);
+	}
+
+	getActionData(currentAccount: string, action: ActivityAction): ActionData {
+		const data = super.getActionData(currentAccount, action);
+		const act = String(action.action);
+		let explanation;
+		if ('delegatebw' === act) {
+			explanation = `${action.data['from']} delegated NET: ${action.data['stake_net_quantity']} & CPU: ${action.data['stake_cpu_quantity']} to ${action.data['receiver']}`;
```
aaroncox, line 59: I would hold off on all of this, I don't think we're going to use it. 

The plan is to actually design custom elements for known transactions, not to explain them directly like this.

I don't want you to end up using a lot of your time for something we'd likely remove 😅

ttwishing: @aaroncox please review again. I changed the loading method, I'm not sure if this is the right thing to do

aaroncox (approved):

PR #206, opened 2024-11-13
Style updates

deansallinen:
Fixes grey usage across the site
Updates spacing and layout for side nav, account switcher, layout header, page header
Centers text on early access banner


tailwind.config.js
```diff
@@ -1,23 +1,21 @@
...
-		// 	lg: ['0.875rem', '1.25rem'], // 14px
-		// 	xl: ['1rem', '1.5rem'], // 16px
-		// 	'2xl': ['1.125rem', '1.75rem'], // 18px
-		// 	'3xl': ['1.25rem', '1.75rem'], // 20px
-		// 	'4xl': ['1.5rem', '2rem'], // 24px
-		// 	'5xl': ['1.875rem', '2.25rem'], // 30px
-		// 	'6xl': ['2.25rem', '2.5rem'], // 36px
-		// 	'7xl': ['3rem', '1'] // 48px
-		// },
+		colors: {
```
dafuga, line 9: Aren't all these colours already there by default? 🤔

deansallinen: Yup, but by importing and defining only the ones we're using, it makes it so the others are unavailable.

That way, no one can use `bg-gray-900` or `bg-slate-900` when they should be using `bg-mineShaft-900`, for example.

PR #208, opened 2024-11-13
RAM pages touchups

dafuga:


src/routes/[network]/(account)/ram/+page.svelte
```diff
@@ -64,7 +64,7 @@
 	let bytesInput: BytesInput;
 </script>
 
-<div class="grid gap-6 lg:grid-cols-2">
+<div class="grid gap-6 pb-6 lg:grid-cols-2">
```
Comment (deansallinen) [inline src/routes/[network]/(account)/ram/+page.svelte:67]: I'm adding bottom padding at the layout level in my latest style-update branch. I assume that's why this is here

src/routes/[network]/(account)/ram/+page.svelte
```diff
@@ -190,22 +190,24 @@
...
+	<Stack class="gap-6 lg:col-start-2">
+		<Grid class="gap-6">
+			<Card>
+				<h3 class="text-muted">
+					RAM Market Cap {data.network.chain.systemToken?.symbol.code || ''}
+				</h3>
+				<AssetText variant="full" value={marketCapEOS} class="text-right" />
+			</Card>
+
+			<Card>
```
Comment (deansallinen) [inline src/routes/[network]/(account)/ram/+page.svelte:202]: This reverts a change I made. I guess it's a matter of preference, but I intended for the third card to NOT span across the bottom since there are breakpoints where it appears too long. 

Example:
Screen Shot 2024-11-12 at 17 51 53

The description is too far removed from the data. I found it was a better UX to deviate from the figma design and keep even width cards instead of having it span. 

Screen Shot 2024-11-12 at 17 53 36

Comment (dafuga) [reply src/routes/[network]/(account)/ram/+page.svelte:202]: Hmm your second screenshot makes sense to me, but that's not what I was seeing on dev. At least on some screen sizes, the third card was on the next row and the fact that it was sticking to the left side of the parent div made it look a bit weird 😅 Now that I understand what you were going for let me see if I can just fix that 😄

src/routes/[network]/(account)/ram/+page.svelte
Comment (dafuga) [reply src/routes/[network]/(account)/ram/+page.svelte:67]: Ah perfect 👍

src/routes/[network]/(account)/ram/+page.svelte
Comment (deansallinen) [reply src/routes/[network]/(account)/ram/+page.svelte:202]: No what you're seeing was correct, when the screen size is small enough it'll make a second row and "stick to the left". I think having it left aligned rather than stretched makes more sense. Until we get a 4th data point or a REALLY large market cap, lol

PR #209, opened 2024-11-13
Fix AssetText format exception when asset.value exceeding 53 bits

ttwishing:
https://v2.unicove.com/eos/api/account/eosio/balances

{
  contract: 'tttingenesis',
  amount: '288144141.51524401',
  decimals: '8',
  currency: 'TTT'
}
Asset.from(`288144141.51524401 TTT`);


aaroncox:
Would this would still cause an error if the number exceeded 53 bits?

The quantity value is a string, but it's immediately being cast to a JS number in the same line, which would cause it to throw the error... unless I'm misunderstanding 🤔

ttwishing:
> Would this would still cause an error if the number exceeded 53 bits?
> 
> The quantity value is a string, but it's immediately being cast to a JS number in the same line, which would cause it to throw the error... unless I'm misunderstanding 🤔

Yes, that's my understanding. But no exception when used quantity. I will study it again.

ttwishing:
Number(''1288144141.51524411') will lose precision when the conversion process exceeds the number of significant digits, and the output will be '288,144,141.515244'. My solution is imperfect, I will correct it again

aaroncox:
Here's an alternative pull request using Big.js, which is the library we're using for the input values.

https://github.com/greymass/2nicove/pull/225

aaroncox:
Implemented the other solution for the time being, even if we lose slight precision during math operations. We can address it in future PRs if the inaccuracy becomes a problem.

PR #225, opened 2024-11-16
Use Big.js to safely wrap asset quantities

aaroncox:


ttwishing:
{
    "currency": "TTT",
    "decimals": "8",
    "contract": "tttingenesis",
    "amount": "288144141.51524401"
}

the output is: 288,144,141.51524400

The last digit of the decimal place may be lost.

PR #236, opened 2024-11-18
enhancement: account loading performance

deansallinen:
Sends a promise to the client to resolve rather than blocking rendering. This allows us to use the native loading state built into sveltekit and stream data when it's available to render.

### DIFF
--- a/src/routes/[network]/(explorer)/account/[name]/+layout.ts
+++ b/src/routes/[network]/(explorer)/account/[name]/+layout.ts
@@ -1,27 +1,27 @@
 import { AccountState } from '$lib/state/client/account.svelte';
-import { error, type LoadEvent } from '@sveltejs/kit';
 import { Name } from '@wharfkit/antelope';
+import type { LayoutLoad } from './$types';
 
-export const load = async ({ fetch, params, parent }: LoadEvent) => {
+// TODO: find the appropriate map for this
+const chainNames = {
+	eos: 'EOS'
+};
+
+export const load: LayoutLoad = async ({ fetch, params, parent }) => {
+	const formattedNetworkName = chainNames[params.network as keyof typeof chainNames];
+
+	// Always call awaited functions last to avoid delayed render
 	const { network } = await parent();
-	let account: AccountState;
-	try {
-		account = await AccountState.for(network, Name.from(String(params.name)), fetch);
-	} catch (e) {
-		console.error(e);
-		error(404, {
-			message: `Account not found: ${String(params.name)}`,
-			code: 'NOT_FOUND'
-		});
-	}
+	const account = AccountState.for(network, Name.from(String(params.name)), fetch);
+
 	return {
 		account,
-		name: account.name,
-		title: `${account.name}`,
-		subtitle: `Account overview on the ${network.chain.name} Network`,
+		name: params.name,
+		title: params.name,
+		subtitle: `Account overview on the ${formattedNetworkName} Network`,
 		pageMetaTags: {
-			title: `${account.name} | ${network.chain.name} Network Account`,
-			description: `An overview of the ${account.name} account on the ${network.chain.name} Network. View account assets, activity, resources and more.`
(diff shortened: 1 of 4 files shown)


src/routes/[network]/(explorer)/account/[name]/+layout.ts
```diff
@@ -1,27 +1,27 @@
 import { AccountState } from '$lib/state/client/account.svelte';
-import { error, type LoadEvent } from '@sveltejs/kit';
 import { Name } from '@wharfkit/antelope';
+import type { LayoutLoad } from './$types';
 
-export const load = async ({ fetch, params, parent }: LoadEvent) => {
+// TODO: find the appropriate map for this
+const chainNames = {
+	eos: 'EOS'
+};
```
Comment (deansallinen) [inline src/routes/[network]/(explorer)/account/[name]/+layout.ts:8]: I'm sure these chain names are formatted somewhere. If we use a lookup table like this for the metadata rather than `const { network } = await parent()` we should reduce render blocking, even if just by a little bit.

src/routes/[network]/(explorer)/account/[name]/+layout.ts
```diff
@@ -1,27 +1,27 @@
...
+
+export const load: LayoutLoad = async ({ fetch, params, parent }) => {
+	const formattedNetworkName = chainNames[params.network as keyof typeof chainNames];
```
Comment (deansallinen) [inline src/routes/[network]/(explorer)/account/[name]/+layout.ts:11]: use params.network to index the table of formatted chain names.

src/routes/[network]/(explorer)/account/[name]/+layout.ts
```diff
@@ -1,27 +1,27 @@
...
-		title: `${account.name}`,
-		subtitle: `Account overview on the ${network.chain.name} Network`,
+		name: params.name,
+		title: params.name,
+		subtitle: `Account overview on the ${formattedNetworkName} Network`,
 		pageMetaTags: {
-			title: `${account.name} | ${network.chain.name} Network Account`,
-			description: `An overview of the ${account.name} account on the ${network.chain.name} Network. View account assets, activity, resources and more.`
+			title: `${params.name} | ${formattedNetworkName} Network Account`,
+			description: `An overview of the ${params.name} account on the ${formattedNetworkName} Network. View account assets, activity, resources and more.`
```
Comment (deansallinen) [inline src/routes/[network]/(explorer)/account/[name]/+layout.ts:24]: modified all metatdata to use the params or the lookup table, this way the only thing being `await`ed is the account.

src/routes/[network]/(explorer)/account/[name]/+page.svelte
```diff
@@ -52,8 +25,11 @@
 	</td>
 {/snippet}
 
-<!-- What gets shown on this page if data.account doesn't exist? -->
```
Comment (deansallinen) [inline src/routes/[network]/(explorer)/account/[name]/+page.svelte:55]: This question is now answered!

src/routes/[network]/(explorer)/account/[name]/+page.svelte
```diff
@@ -52,8 +25,11 @@
...
-{#if data.account}
+{#await data.account}
+	{console.log('loading state hit')}
+	<div class="bg-red-300 text-4xl text-black">Loading...</div>
```
Comment (deansallinen) [inline src/routes/[network]/(explorer)/account/[name]/+page.svelte:30]: Loading state: Here I've intentionally left a big red banner so it's obvious when we're loading. The time we see the red banner is the time we would have spent blocking render.

src/routes/[network]/(explorer)/account/[name]/+page.svelte
```diff
@@ -192,19 +172,21 @@
...
 				{#if isCurrentUser}
 					<Button href={`/${data.network}/resources`} variant="secondary">Go to Resources</Button>
 				{/if}
 			</Card>
 		{/if}
 	</PageColumns>
-{/if}
+{:catch error}
+	{error.message}
+{/await}
```
Comment (deansallinen) [inline src/routes/[network]/(explorer)/account/[name]/+page.svelte:192]: Here we will do our "No account found" state

deansallinen:
Messes with SSR, so not the direction we want to go with this page

PR #247, opened 2024-11-20
Logo updates

deansallinen:
- Use the current network logo on homepage (e.g. Jungle 4, EOS)
- Refactor preloading to happen in layout instead of components
- Preload current chain logo, prefetch the others
- Use the `chainLogos` version of the EOS logo elsewhere to take advantage of the cached image

### DIFF
--- a/src/routes/[network]/(homepage)/(wallets)/metamask/+page.svelte
+++ b/src/routes/[network]/(homepage)/(wallets)/metamask/+page.svelte
@@ -6,13 +6,13 @@
 	import Button from '$lib/components/button/button.svelte';
 	import Box from '$lib/components/layout/box/box.svelte';
 	import { checkForSnap, requestSnap } from '$lib/metamask-snap';
-	import EOS from '$lib/assets/EOS@2x.svg';
 	import Metamask from '$lib/assets/metamask.svg';
 	import { MetaMaskState } from '$lib/state/metamask.svelte';
 	import type { UnicoveContext } from '$lib/state/client.svelte.js';
 	import { Cluster, Stack } from '$lib/components/layout/index.js';
 	import { accountCreationPluginMetamask } from '$lib/state/client/wharf.svelte';
 	import { getSetting } from '$lib/state/settings.svelte';
+	import { chainLogos } from '@wharfkit/common';
 
 	const { data } = $props();
 	const context = getContext<UnicoveContext>('state');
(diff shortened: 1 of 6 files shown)


src/routes/[network]/(homepage)/(wallets)/metamask/+page.svelte
```diff
@@ -92,6 +92,8 @@
 			walletPlugin: 'wallet-plugin-metamask'
 		});
 	}
+
+	const EOS = String(chainLogos.get(String(data.network.chain.id)));
```
Comment (aaroncox) [inline src/routes/[network]/(homepage)/(wallets)/metamask/+page.svelte:96]: Should be `logo` or something, since it's not always EOS? 🤔

src/routes/[network]/(homepage)/+page.svelte
```diff
@@ -158,7 +160,11 @@
 					<svg width="36" height="36" fill="none" xmlns="http://www.w3.org/2000/svg">
 						<path d="M18.008 0v36M36.008 18h-36" stroke="#fff" />
 					</svg>
-					<img class="h-40 rounded-full bg-mineShaft-950 px-2 py-4" src={EOS} alt="eos" />
+					<img
+						class="h-40 rounded-full bg-mineShaft-950 object-contain px-2 py-4"
+						src={String(chainLogos.get(String(Chains.EOS.id)))}
```
Comment (aaroncox) [inline src/routes/[network]/(homepage)/+page.svelte:165]: If this is an EOS specific homepage this makes sense, but if its some other network - should this be based off the network params?

src/routes/[network]/(homepage)/components/tlvhex.svelte
```diff
@@ -21,7 +27,7 @@
 			/>
 		</svg>
 		<div class="grid text-center uppercase">
-			<img class="mb-4 h-12 place-self-center md:h-20" src={EOS} alt="eos" />
+			<img class="mb-4 h-12 place-self-center md:h-20" src={networkLogo} alt="eos" />
```
Comment (aaroncox) [inline src/routes/[network]/(homepage)/components/tlvhex.svelte:30]: `alt="eos"` should match current network name

src/routes/[network]/(homepage)/+page.svelte
Comment (deansallinen) [reply src/routes/[network]/(homepage)/+page.svelte:165]: Hmm... I thought the metamask integration was only for EOS? If not, I can make the change

src/routes/[network]/(homepage)/components/tlvhex.svelte
Comment (deansallinen) [reply src/routes/[network]/(homepage)/components/tlvhex.svelte:30]: Good catch, thanks.

src/routes/[network]/(homepage)/+page.svelte
Comment (aaroncox) [reply src/routes/[network]/(homepage)/+page.svelte]: To add - on Unicove it should work for both Jungle 4 and Kylin

PR #251, opened 2024-11-20
dev: update interval in env

deansallinen:
If you set `PUBLIC_ACCOUNT_UPDATE_INTERVAL` and `PUBLIC_NETWORK_UPDATE_INTERVAL` in your `.env` to another value, it will override the default values and prevent hitting the rate limit on the internal api when iterating quickly with multiple clients


deansallinen:
Build is failing because the key doesn't exist in the CI env

PR #259, opened 2024-11-22
Key page suggestions

deansallinen:
See comments


src/routes/[network]/(explorer)/key/[publicKey]/+page.svelte
```diff
@@ -1,40 +1,46 @@
...
-									{m.account_page()}
-									<ArrowRight size={20} />
-								</span>
-							</a>
-						</dd>
-					</div>
-				{/each}
+				<!-- TODO: Display publicKey in multiple formats -->
+				<p>{data.publicKey}</p>
+				<p>{data.publicKey}</p>
```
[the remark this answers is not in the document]
Comment (aaroncox) [reply src/routes/[network]/(explorer)/key/[publicKey]/+page.svelte:19]: Yeah for sure.

```ts
const pubkey = PublicKey.from(data.publicKey)

console.log(String(pubkey))
// PUB_K1_6LPN8xGye86wme4PwqN8TSxAfGTQvqBcN5jGnURJPHyqn3a973

console.log(pubkey.toLegacyString())
// EOS6LPN8xGye86wme4PwqN8TSxAfGTQvqBcN5jGnURJPHyqnXqNAR
```

src/lib/components/elements/account.svelte
```diff
@@ -0,0 +1,80 @@
...
+		children?: Snippet;
+	}
+
+	let { name, contract = false, children, ...props }: Props = $props();
+
+	const path = contract ? '/contract/' + name : '/account/' + name;
+
+	let { network } = getContext<UnicoveContext>('state');
+
+	let account: AccountState | undefined = $state();
```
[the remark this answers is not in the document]
dafuga: We could have some kind of account cache store that is used to accomplish this, but we would probably always end up doing an API request since some of the resource values could be updated at any time. That store could immediately return an old version of the account data while the new data is loaded though which would allow the page load to be instant in this case.

src/lib/components/elements/account.svelte
```diff
@@ -0,0 +1,84 @@
...
+>
+	{#if children}
+		{@render children()}
+	{:else}
+		{name}
+	{/if}
+</a>
+
+{#if $open && account}
+	<div
```
dafuga, line 63: This is a great idea! I think we should probably make a Preview component that would make it easier to have similar previews for other data types (for example, it would be great to have a transaction preview).

src/routes/[network]/(explorer)/key/[publicKey]/+page.ts
```diff
@@ -9,11 +10,12 @@ export const load: Load = async ({ fetch, params, parent }) => {
 
 	const title = m.key_page_title();
 	const description = m.key_page_description();
+	const pubkey = PublicKey.from(String(params.publicKey));
```
Comment (dafuga) [inline src/routes/[network]/(explorer)/key/[publicKey]/+page.ts:13]: This is going to break the page when an invalid key is used as a param. I think we either need to just pass the param down as a prop or we need to catch PublicKey errors here.

src/routes/[network]/(explorer)/key/[publicKey]/+page.svelte
```diff
@@ -1,40 +1,49 @@
...
-							<a href="/{data.network}/account/{account}" class="text-blue-600 hover:text-blue-800">
-								<span class="inline-flex items-center gap-2">
-									{m.account_page()}
-									<ArrowRight size={20} />
-								</span>
-							</a>
-						</dd>
-					</div>
-				{/each}
+<MultiCard>
```
Comment (dafuga) [inline src/routes/[network]/(explorer)/key/[publicKey]/+page.svelte:13]: ❤️

src/routes/[network]/(explorer)/key/[publicKey]/+page.svelte
```diff
@@ -1,40 +1,49 @@
...
+		{:then accounts}
+			{#if accounts && accounts.length > 0}
+				<Grid tag="ul">
+					{#each accounts as account}
+						<li class="flex">
+							<Account
+								name={account}
+								class="h-12 content-center rounded-lg px-4 hover:bg-mineShaft-900 hover:text-mineShaft-50 focus-visible:bg-mineShaft-900 focus-visible:text-mineShaft-50"
+							/>
+							<!-- <a href={`/eos/account/${account}`}>{account}</a> -->
```
Comment (dafuga) [inline src/routes/[network]/(explorer)/key/[publicKey]/+page.svelte:38]: I think we can just remove this.

dafuga (changes requested):
I left you a few comments, but the main thing is that we need to handle the user passing an invalid key (right now it's causing a 500 error).

src/lib/components/elements/account.svelte
deansallinen: I don't know how we would accomplish it with a generic component, but we can definitely do the same pattern with the `Transaction` component. Just need an idea of what to preview!

src/routes/[network]/(explorer)/key/[publicKey]/+page.ts
Comment (deansallinen) [reply src/routes/[network]/(explorer)/key/[publicKey]/+page.ts:13]: I think we _should_ be breaking the page on invalid key and sending the user to the `+error.svelte` 404 page. 

I don't know how we'd handle it otherwise. We can't really show anything else on the page if there's no key.

Comment (dafuga) [reply src/routes/[network]/(explorer)/key/[publicKey]/+page.ts:13]: From the point of view of the user, I think the way that it's being handled now that you added the 404 page is fine, but I don't think it's a good habit to let pages break like that (even if the user is taken to the 404 page). I would add a try catch block around the ` PublicKey.from(String(params.publicKey));` that sends users to the 404 page when an invalid key format error occurs. At some point, we're probably going to want to start monitoring our error logs and the less noise that is in there the better.

src/lib/components/elements/account.svelte
dafuga: Either we would pass a trigger snippet that would allow you display the preview (passed as children prop) or the other thing we could do is have a generic preview component that would just take the data (in this case the account data) as a prop and the component would already know how to display the different data types (account, transaction, etc..). In any case, I don't think that we need to worry about this now.

src/routes/[network]/(explorer)/key/[publicKey]/+page.ts
```diff
@@ -26,7 +26,6 @@ export const load: PageLoad = async ({ fetch, params, parent }) => {
 
 	return {
 		title,
-		subtitle: `${params.publicKey?.slice(0, 10)}...${params.publicKey?.slice(-10)}`,
```
Comment (deansallinen) [inline src/routes/[network]/(explorer)/key/[publicKey]/+page.ts:15]: I've removed it from here since we also have it just below in a card... doesn't make much sense to have a truncated version so close in the UI to the full version

src/routes/[network]/(explorer)/key/[publicKey]/+page.ts
Comment (deansallinen) [reply src/routes/[network]/(explorer)/key/[publicKey]/+page.ts]: I've made the error handling explicit. It's a good point, because this way we can switch on the error code to show different components in the `+error.svelte`

dafuga (approved):

PR #275, opened 2024-11-28
design block page

ttwishing:


aaroncox:
It's looking good. Here are a few changes I'll request:

---

image

In the header here for the timestamp, lets show the milliseconds, since there's 2 blocks per second.

---

image

Reduce the font size on the block ID to prevent a 3rd line of text.

deansallinen:
> Reduce the font size on the block ID to prevent a 3rd line of text.

We shouldn't be arbitrarily reducing text sizes in elements to match layout preferences. Instead, in this case, we should be using `text-balance` to prevent orphaned characters.

I'm making a PR to this branch which will include this change and other styling updates.

PR #277, opened 2024-11-28
Dean block page

deansallinen:
- changed markup to use common components
- updated description list component to receive children
- combined and moved data processing to `layout.ts` 

I took some design liberties as follows:
- used a description list pattern instead of a table pattern for the block details. Reason: this is key-value data, not tabular data, and we have this pattern elsewhere so I'm changing it for consistency.
- right aligned numerical data
- Use `TransactionText` and `AccountText` components where applicable

### DIFF
--- a/src/routes/[network]/(explorer)/block/[number]/+page.svelte
+++ b/src/routes/[network]/(explorer)/block/[number]/+page.svelte
@@ -7,7 +7,14 @@
 	import ChevronRight from 'lucide-svelte/icons/chevron-right';
 	import { truncateCenter } from '$lib/utils';
 
-	const { data } = $props();
+	let { data } = $props();
+
+	const details = $derived([
+		{ key: 'Block Number', value: data.block.block_num },
+		{ key: 'Producer Name', value: data.block.producer },
+		{ key: 'Total Actions', value: data.actions },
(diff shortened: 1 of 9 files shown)


src/routes/[network]/(explorer)/block/[number]/+page.svelte
```diff
@@ -1,19 +1,82 @@
...
+					</tr>
+				</thead>
+				<tbody>
+					{#each transactions as transaction}
+						{#if transaction}
+							<tr>
+								<td><TransactionText id={transaction.trx.id} /></td>
+								<td class="text-right">{transaction.trx.transaction?.actions.length}</td>
+								<td class="text-right">{transaction.cpu_usage_us}</td>
+								<td class="text-right">{transaction.net_usage_words}</td>
```
Comment (aaroncox) [inline src/routes/[network]/(explorer)/block/[number]/+page.svelte:38]: Needs the `* 8` to multiply words into bytes.

Comment (deansallinen) [reply src/routes/[network]/(explorer)/block/[number]/+page.svelte:38]: Thanks. I'll make the change tomorrow and move the strings to the translation file.

deansallinen:
Fixed the issues with the diff not displaying correctly. Now we can see what's changed

PR #307, opened 2024-12-12
Permissions page

deansallinen:


src/routes/[network]/(explorer)/account/[name]/permissions/permission.svelte
```diff
@@ -0,0 +1,154 @@
...
+	import { Clock } from 'lucide-svelte';
+	dayjs.extend(duration);
+	dayjs.extend(relativeTime);
+	interface Props {
+		permission: TreePermission;
+		level?: number;
+	}
+	let { level = 0, ...props }: Props = $props();
+	let { permission, children } = $derived(props.permission);
+	$inspect(permission);
```
Comment (aaroncox) [inline src/routes/[network]/(explorer)/account/[name]/permissions/permission.svelte:20]: Maybe enable if debugMode is true? Either that or we can remove.

src/routes/[network]/(explorer)/account/[name]/permissions/permission.svelte
```diff
@@ -0,0 +1,154 @@
...
+	import type { TreePermission } from './+page';
+	import Key from '$lib/components/elements/key.svelte';
+	import CopyButton from '$lib/components/button/copy.svelte';
+	import Account from '$lib/components/elements/account.svelte';
+	import Contract from '$lib/components/elements/contract.svelte';
+	import dayjs from 'dayjs';
+	import duration from 'dayjs/plugin/duration'; // ES 2015
+	import relativeTime from 'dayjs/plugin/relativeTime'; // ES 2015
+	import { Clock } from 'lucide-svelte';
+	dayjs.extend(duration);
```
Comment (aaroncox) [inline src/routes/[network]/(explorer)/account/[name]/permissions/permission.svelte:12]: Is there a way we could set these globally, or could we do it in some other way? 

There's a couple pages now that are doing this and it might get repetitive 🤔

aaroncox (approved):

src/routes/[network]/(explorer)/account/[name]/permissions/permission.svelte
Comment (deansallinen) [reply src/routes/[network]/(explorer)/account/[name]/permissions/permission.svelte]: Removed

src/routes/[network]/(explorer)/account/[name]/permissions/permission.svelte
Comment (deansallinen) [reply src/routes/[network]/(explorer)/account/[name]/permissions/permission.svelte]: Done - refactored to setup dayjs once globally and you can just `import dayjs from 'dayjs'` and it'll include all the relevant plugins

PR #318, opened 2024-12-18
Funding page

dafuga:


src/routes/[network]/(account)/fund/+page.svelte
```diff
@@ -59,49 +122,102 @@
...
-	<li><a href="https://www.coinbase.com">https://www.coinbase.com</a></li>
-	<li><a href="https://www.kraken.com">https://www.kraken.com</a></li>
-	<li><a href="https://www.okx.com">https://www.okx.com</a></li>
-	<li><a href="https://global-aws.huobi.com">https://global-aws.huobi.com</a></li>
-	<li><a href="https://www.gate.io">https://www.gate.io</a></li>
-	<li><a href="https://upbit.com">https://upbit.com</a></li>
-	<li><a href="https://kucoin.com">https://kucoin.com</a></li>
-</ul>
+
+<MultiCard class="grid grid-cols-1 gap-4 sm:grid-cols-2 lg:grid-cols-4">
```
Comment (deansallinen) [inline src/routes/[network]/(account)/fund/+page.svelte:194]: If we're overriding the display context anyways, I think the `Grid` layout component would be better here. Equal width items that will fill available space and can get rid of the media queries too.

```suggestion
<Grid>
```

src/routes/[network]/(account)/fund/+page.svelte
```diff
@@ -59,49 +122,102 @@
...
+				rel="noopener noreferrer"
+				class="block h-full transition-all hover:scale-105"
+			>
+				<div class="relative mb-4 aspect-square overflow-hidden rounded-lg bg-mineShaft-900 p-4">
+					<img
+						src={exchange.logo}
+						alt="{exchange.name} logo"
+						class="absolute left-1/2 top-1/2 h-full w-full -translate-x-1/2 -translate-y-1/2 transform object-cover p-4"
+					/>
+				</div>
```
Comment (deansallinen) [inline src/routes/[network]/(account)/fund/+page.svelte:209]: All this centering can be replaced with:

```suggestion
				<div class="mb-4 aspect-square grid place-items-center rounded-lg bg-mineShaft-900 p-4">
					<img
						src={exchange.logo}
						alt="{exchange.name} logo"
						class="size-full object-cover p-4"
					/>
				</div>
```

static/binance.webp
deansallinen, line 1: Goes for all the images:

We should put these in the `src/lib/assets` directory, then import them in the `<script>` tag of the component. That way Vite can preprocess them. Also, in my branch for the homepage I'm adding the `@svelte/enhanced-img` plugin which will take it one step further and generate a `srcset` for each image when using `<enhanced:img>` tag.

We can upgrade to the enhanced image tag later, but for now I'd say let's put the images in a folder in `assets`

src/routes/[network]/(account)/fund/+page.svelte
```diff
@@ -59,49 +122,102 @@
...
-{#if !context.account}
-	<p>You must be logged in with an account to use this feature.</p>
-{:else if !coinbaseInstance}
-	<p>No supported funding methods for this blockchain.</p>
-{:else}
-	<Button onclick={coinbase}>Buy EOS with Coinbase</Button>
-{/if}
+<h4 class="h4">Token Purchase Providers:</h4>
+
+<MultiCard class="max-w-[700px]">
```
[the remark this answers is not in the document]
Comment (deansallinen) [reply src/routes/[network]/(account)/fund/+page.svelte:147]: Removing it enlarges the card. Maybe we want to set a `max-w` on the Card, or use a `Grid` layout component and set `itemWidth='someValue'`

src/routes/[network]/(account)/fund/+page.svelte
```diff
@@ -1,15 +1,79 @@
...
+				range: '1.49% - 3.99%'
+			},
+			limits: {
+				daily: m._500_per_day()
+			},
+			details: [m.buy_or_transfer_crypto(), m.debit_and_credit_cards_supported()],
+			action: {
+				text: m.buy_eos_with_coinbase(),
+				handler: 'coinbase',
+				className: 'bg-blue-600 hover:bg-blue-500'
```
Comment (deansallinen) [inline src/routes/[network]/(account)/fund/+page.svelte:26]: Why are we overriding the Button classes?

src/routes/[network]/(account)/fund/+page.svelte
```diff
@@ -59,49 +122,102 @@
...
-	<Button onclick={coinbase}>Buy EOS with Coinbase</Button>
-{/if}
+<h4 class="h4">Token Purchase Providers:</h4>
+
+<MultiCard class="max-w-[700px]">
+	{#each ON_RAMP_PROVIDERS as service}
+		<Card>
+			<div>
+				<div class="mb-4 flex items-center justify-center">
+					<img src={service.logo} alt={service.id} class="h-24 w-3/5 object-contain" />
```
Comment (deansallinen) [inline src/routes/[network]/(account)/fund/+page.svelte:152]: `w-3/5` seems pretty specific to the coinbase logo, but we can update that when we have more cards

messages/en.json
```diff
@@ -90,17 +94,26 @@
 	"contract_tables_view_title": "Table: {table}",
 	"contract_view_description": "An overview of the {contract} smart contract on the {network} network. This contract contains {actions} actions, {tables} tables, and {structs} structs.",
 	"contract_view_title": "Contract: {contract} | {network}",
+	"debit_and_credit_cards_supported": "Debit and credit cards supported",
 	"delegation_metadata_refund_description": "Claim previously delegated {network} tokens.",
 	"delegation_metadata_refund_title": "Refund Delegated {network} Tokens",
+	"details": "Details",
 	"error_loading_accounts": "Error loading accounts: {error}",
+	"exchanges": "Exchanges:",
```
deansallinen, line 102:
```suggestion
	"exchanges": "Exchanges",
```

src/routes/[network]/(account)/fund/+page.svelte
Comment (deansallinen) [reply src/routes/[network]/(account)/fund/+page.svelte:194]: Looks like there's a little bug with the `<Grid>` component right now, but I'll make a fix. In the meantime `<Grid tag='div'> will work.

With the fix, I'll make it so you can do `<Grid tag='ul'>` and update the `Card` to take a tag too, so it can be a `li`

deansallinen (commented):
Some changes, let me know if you'd rather I make a PR

src/routes/[network]/(account)/fund/+page.svelte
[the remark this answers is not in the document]
Comment (dafuga) [reply src/routes/[network]/(account)/fund/+page.svelte:147]: I tried adding the max width on the card itself, but it would add a gap between cards on larger screens which is not what we want. Grid may be what I need though, thanks!

src/routes/[network]/(account)/fund/+page.svelte
Comment (dafuga) [reply src/routes/[network]/(account)/fund/+page.svelte:209]: Oh awesome!

src/routes/[network]/(account)/fund/+page.svelte
Comment (dafuga) [reply src/routes/[network]/(account)/fund/+page.svelte:152]: Yeah,we may end up having to make that a configurable value, but I think it's fine for now.

src/routes/[network]/(account)/fund/+page.svelte
Comment (dafuga) [reply src/routes/[network]/(account)/fund/+page.svelte:26]: I wanted to add the ability to customize the button for different providers, but since it's not doing anything I think we can just remove this for now.

static/binance.webp
deansallinen: If you rebase on `dev` you should now be able to use `<enhanced:img>` for these assets

https://svelte.dev/docs/kit/images#sveltejs-enhanced-img

src/routes/[network]/(account)/fund/+page.svelte
Comment (deansallinen) [reply src/routes/[network]/(account)/fund/+page.svelte:194]: `<Grid tag='ul'>` and `<Card tag='li'>` should now work after rebasing to `dev`

PR #319, opened 2024-12-18
typo fixes on the homepage

sandroguras:


aaroncox:
Dean caught one of these in his PR it looks like, and the other is probably fine for now since we may want to rewrite that anyways.

PR #334, opened 2025-01-15
Account activity loading enhancement

deansallinen:


src/routes/[network]/api/account/[[name]]/activity/[[start]]/+server.ts
```diff
@@ -20,12 +20,14 @@ export async function GET({ fetch, params }: RequestEvent) {
 		return json({ error: `Activity not supported on ${network.chain.name}.` }, { status: 500 });
 	}
 
-	const start = Number(params.start) || 1;
-	const requests = [getActivity(network.client, params.name, start)];
-	const headers = getCacheHeaders(5);
+	const start = params.start === '0' ? 1 : Number(params.start);
+
+	// Aggressively cache older activity
```
Comment (aaroncox) [inline src/routes/[network]/api/account/[[name]]/activity/[[start]]/+server.ts:25]: We can't do this, because older activity shifts in time and the results aren't in the same sequence because its in descending order.

If the initial page shows records 99 through 90 (walking back in time), and page 2 shows 89 to 80 - then if a new entry is added to the front, record 90 would end up on page 2. If page 2 is cached, that record won't show.

aaroncox (approved):

PR #335, opened 2025-01-16
Localized Resources pages

ttwishing:


aaroncox:
Looks good, thanks for doing this! Will merge this in now.

I'll add the machine translated Chinese/Korean languages once its merged in.

PR #340, opened 2025-01-21
feat: collapse side menu

deansallinen:
Enable a setting to collapse the side menu on larger screens. Perfect for hot-key-only navigation


deansallinen:
Currently not required

PR #345, opened 2025-01-23
Group notification traces to deduplicate on the transaction page

### LINKED ISSUE #343: Group notification traces to deduplicate on the transaction page
An example transaction: 

https://unicove.com/en/jungle4/transaction/60463e3cad1d3d8185d3398fa001015b19ab57ebc8667a0fc005de345ebcddcb/traces

This currently displays 5 traces, but could be grouped to deduplicate the amount of data shown.

Example grouping of the same transaction: 

Image

Removing the duplicates helps illustrate what actions and inline actions were performed, without listing them out individually and making it appear as if there's multiple transfers.

duyluong1994:
Resolves #343


aaroncox:
Awesome, thank you for the contribution!

I figured there was an elegant way to do this, but just needed to record the idea based on a conversation in Telegram yesterday. I very much appreciate you taking the time to tackle it 🙏

PR #350, opened 2025-02-01
Add icon to page header if an account is a contract #348

### LINKED ISSUE #348: Add icon to page header if an account is a contract
While viewing an account:

https://dev.unicove.com/en/jungle4/account/shipload.gm

If the account has `account.contract` set to `true`, which was added here:

https://github.com/greymass/unicove/pull/344/commits/4ac904f220932a2b70b9753606c3b634f49ca869

We should display an icon at the copy next to the copy button, using the `<Code />` icon, that lets you swap to the alternative contract URL for that account.

https://dev.unicove.com/en/jungle4/contract/shipload.gm

duyluong1994:
Resolve #348

### DIFF
--- a/src/lib/components/button/code.svelte
+++ b/src/lib/components/button/code.svelte
@@ -1,0 +1,41 @@
+<script lang="ts">
+	import { browser } from '$app/environment';
+	import { goto } from '$app/navigation';
+	import type { UnicoveContext } from '$lib/state/client.svelte';
+	import { cn } from '$lib/utils';
+	import Code from 'lucide-svelte/icons/code';
+	import { getContext } from 'svelte';
+
+	const context = getContext<UnicoveContext>('state');
+
+	interface Props {
+		data: string;
+		slop?: boolean;
+	}
+
+	let { slop = true, ...props }: Props = $props();
+
+	function goToContract() {
+		goto(`/${props.data}`);
+	}
+
+	let buttonSize = $derived(slop ? 'size-12' : 'size-4');
+</script>
+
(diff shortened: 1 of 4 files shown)


src/lib/components/button/code.svelte
```diff
@@ -0,0 +1,41 @@
...
+	>
+		<button
+			onclick={goToContract}
+			class={cn(
+				'peer absolute left-1/2 top-1/2 -translate-x-1/2 -translate-y-1/2 focus-visible:outline-none',
+				buttonSize
+			)}
+		>
+			<!-- Button is done this way with absolute positioning so we can maintain a decent hit slop on mobile without affecting layout -->
+		</button>
```
deansallinen, line 38: Also should be a link

src/lib/components/pageheader.svelte
```diff
@@ -50,7 +58,13 @@
...
 			{#if routePath === 'account'}
-				<CopyButton data={props.title} />
+				<CopyButton data={props.title} slop={false} />
+				{#if props.contract}
+					<CodeButton data={contractPath} slop={false} />
+				{/if}
+			{/if}
+			{#if routePath === 'contract'}
+				<AccountButton data={accountPath} slop={false} />
 			{/if}
```
deansallinen, line 68: With this layout, the button under the cursor changes depending on the page - we should keep consistency and either add a copy button to the contract page, or swap the order so the switcher button is first.

deansallinen (changes requested):
Thanks for the PR! A couple of small changes can be made, but otherwise it looks good.

aaroncox:
Merged. Opened issue https://github.com/greymass/unicove/issues/372 to track the requested changes.

PR #376, opened 2025-02-19
David/playwright integration

sandroguras:


tests/main-pages-visibility-test.spec.ts
```diff
@@ -0,0 +1,34 @@
+import { test, expect } from '@playwright/test';
+
+test('test', async ({ page }) => {
+	// Login with test account.
+	await page.goto('https://dev.unicove.com/en/jungle4/debug/account');
```
aaroncox, line 5: Isn't this always going to test what's deployed on this domain, and not what this branch is?

PR #384, opened 2025-02-23
Adding support for multisig accounts

aaroncox:


src/lib/components/accountswitch.svelte
```diff
@@ -252,9 +257,14 @@
...
-											{session.actor}@{session.permission}
-										</span>
+										<div class="text-left font-medium">
+											<div>{session.actor}@{session.permission}</div>
+											{#if session.walletPlugin.id === 'wallet-plugin-multisig'}
+												<div class="text-xs">
+													↳ multisig using {session.walletPlugin.data.session.actor}
+												</div>
+											{/if}
+										</div>
```
deansallinen, line 267: I'm curious what this looks like in context with layout, the arrow glyph, and the small text. 

Probably should use a translation here too?

src/lib/components/transact/summary.svelte
```diff
@@ -15,15 +23,44 @@
...
-			</picture>
-			<h2 class="h3">{m.common_transaction_complete()}</h2>
+			{#if proposals && proposals.length}
+				<picture class="size-24">
+					<ClipboardPen class="size-full text-green-300" />
+				</picture>
+				<h3 class="h3">Multi-Sig Proposal Created</h3>
+				<p class="text-center">
+					The multi-sig proposal for this transaction has been created and now needs to be approved.
+					View the proposal below and share it with the parties who need to sign.
```
deansallinen, line 49: Should we be picky about translations at this point?

src/routes/[network]/(explorer)/msig/[proposer]/[proposal]/+page.svelte
```diff
@@ -12,6 +12,7 @@
 
 	import { ApprovalManager } from './manager.svelte';
 	import type { ActionDisplayVariants } from '$lib/types';
+	import { goto } from '$app/navigation';
```
Comment (deansallinen) [inline src/routes/[network]/(explorer)/msig/[proposer]/[proposal]/+page.svelte:15]: Wonder if we should use the wrapped `goto` from `$lib/utils` so the language tag is included

src/routes/[network]/settings/+page.svelte
```diff
@@ -38,6 +49,49 @@
 	$effect(() => {
 		context.settings.data.debugMode = advancedMode ? debugMode : false;
 	});
+
+	const range: ExtendedSelectOption[] = [
+		{ label: '15 mintes', value: TimeSeconds['15m'] },
```
Comment (deansallinen) [inline src/routes/[network]/settings/+page.svelte:54]:
```suggestion
		{ label: '15 minutes', value: TimeSeconds['15m'] },
```

static/_site.webmanifest
deansallinen, line 1: Do we still need to rename this file if we're not linking to it in the `<head>` anymore?

src/routes/[network]/(account)/multisig/+page.svelte
```diff
@@ -0,0 +1,444 @@
...
+				</DD>
+			</DLRow>
+		</DL>
+		<fieldset class="flex items-center gap-3">
+			<Checkbox bind:checked={ramBytesUseCustom} id="rambytes-custom" />
+			<Label for="rambytes-custom">Change RAM options for new account</Label>
+		</fieldset>
+		<fieldset class="grid gap-3">
+			{#if ramBytesUseCustom}
+				<SingleCard>
```
Comment (deansallinen) [inline src/routes/[network]/(account)/multisig/+page.svelte:325]: This component was designed with page-level layout in mind. Will have to evaluate how it works in this context.

src/lib/components/select/select.svelte
```diff
@@ -7,7 +7,7 @@
 
 	interface Props {
 		options: ExtendedSelectOption[];
-		selected: ExtendedSelectOption;
+		selected?: ExtendedSelectOption;
```
deansallinen, line 10: IIRC this was required in the `melt-ui` builder so I made it required in the interface

src/routes/[network]/(explorer)/account/[name]/authority/+page.svelte
```diff
@@ -0,0 +1,62 @@
...
+	The {data.account.name} account has some form of authority assigned to the other accounts listed below.
+</p>
+
+{#await data.authorizations then authorizations}
+	{#each authorizations.accounts as auth}
+		{@const permission = PermissionLevel.from(`${auth.account_name}@${auth.permission_name}`)}
+		<Card>
+			<div class="flex">
+				<div class="flex-1">
+					<Account name={auth.account_name} class="h2">
```
Comment (deansallinen) [inline src/routes/[network]/(explorer)/account/[name]/authority/+page.svelte:38]: Based on the video in Slack, this should probably be smaller than `h2` since it's competing in prominence with the page header. Try `h3` or `h4`

src/routes/[network]/(explorer)/msig/[proposer]/[proposal]/+page.svelte
Comment (aaroncox) [reply src/routes/[network]/(explorer)/msig/[proposer]/[proposal]/+page.svelte:15]: Good call, one less redirect!

Comment (aaroncox) [reply src/routes/[network]/(explorer)/msig/[proposer]/[proposal]/+page.svelte]: Updated a few pages where this made sense.

https://github.com/greymass/unicove/pull/384/commits/6cfe7e9a5d3b6d46fa47df14011a8c7f1d76576f

src/lib/components/accountswitch.svelte
aaroncox: image

Here's what it looks like. I've updated all the translations in the account switcher here: https://github.com/greymass/unicove/pull/384/commits/e69236960f8cfe4fc1309ca29a80afa84085ab46

src/lib/components/transact/summary.svelte
aaroncox: We probably should get better about this, and have someone in charge of that entire pipeline. Considering we're not 100% translated right now though, I don't know if it should be a barrier?

static/_site.webmanifest
aaroncox: No, probably not. I'll undo the rename.

src/routes/[network]/(account)/multisig/+page.svelte
Comment (aaroncox) [reply src/routes/[network]/(account)/multisig/+page.svelte:325]: I'm probably using a bunch of components wrong 😅 

This page in particular shouldn't have even been committed. It is just a copy of one of the other forms, that I never got around to using. I'll delete it for the time being.

src/lib/components/select/select.svelte
aaroncox: I'll revert this change and force a default option on the settings page then. 

The select element with a default option ands up triggering the effects on the page, setting the value to whatever the select believes the default option should be. This could override defaults set elsewhere. 

So there's some funk in here at the moment we should try to figure out how to address.

src/lib/components/accountswitch.svelte
deansallinen: It might call for a slightly lighter-weight font in the second line, but otherwise looks good. 

I'm always hesitant to use glyphs in the font vs an svg because they can become unpredicatable with different fonts, different browser settings, zoom levels, but more importantly: it's decorative, so it shouldn't be included in the text string (which will be translated at some point)

src/lib/components/select/select.svelte
deansallinen: I agree, we should address the funk and see if we can make `selected` optional. It does make sense to have the option to create a dropdown with nothing selected and no default.

deansallinen: https://github.com/greymass/unicove/issues/389

src/lib/components/accountswitch.svelte
aaroncox: Yeah totally understood. We probably want a "real" design for these elements, since as I worked on this I was just implementing my developer design to make something available. 

I'm very open to change with how its displayed.

src/lib/components/select/select.svelte
aaroncox: Awesome, thanks!

PR #409, opened 2025-03-20
feat: vaulta swap info page

deansallinen:


aaroncox:
Can always reopen if we go this route!

PR #438, opened 2025-04-28
refactor: pageheader actions

deansallinen:
Pageheader actions (e.g. copy, switch between account/contract) are now defined in the page data (i.e. +page.ts) so each page can have custom actions or custom data to be copied


src/routes/[network]/(explorer)/account/[name]/+layout.ts
```diff
@@ -30,6 +31,10 @@ export const load: LayoutLoad = async ({ fetch, params }) => {
 		subtitle: m.account_page_subtitle({
 			network: network.chain.name
 		}),
+		header: {
+			copyData: params.name,
+			actions: [{ icon: Code, href: `/${network}/contract/${params.name}` }]
```
Comment (aaroncox) [inline src/routes/[network]/(explorer)/account/[name]/+layout.ts:36]: This is now causing all accounts to have an icon to access the contract, and removes the check that would detect if it had a contract.

PR #472, opened 2025-05-22
feat: button to recollapse expanded data

### LINKED ISSUE #457: Allow toggling of collapsed data (e.g. in action data)
Currently if the data exceeds a certain length, we collapse it by default and offer the ability to expand it. There's however no way to collapse it again aside from refreshing the page.

We should add a way to collapse that data into its minimized state again. Perhaps its another button for "minimize" that is displayed at the bottom of the element, which could also follow the bottom of the viewport to ensure its always collapsible (when viewing pages long information).

deansallinen:
Closes #457


deansallinen:
Will be handled in `unicove-component` library

PR #502, opened 2025-07-02
Disable RentRex for Spring-2-testent

ericpassmore:
Tracked down the `Uncaught (in promise) Error: Invalid number` to misconfigured RentRex on spring2 testnet chain. Disabled RentRex feature and error goes away. 
Screenshot 2025-07-02 at 11 12 53 AM

### DIFF
--- a/src/app.css
+++ b/src/app.css
@@ -191,6 +191,7 @@
 	}
 
 	/* Themes */
+	:root,
 	[data-theme='vaulta'] {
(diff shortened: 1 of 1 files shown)


aaroncox:
Interesting... seems like something with the network state is causing the price to be `NaN`. 

Disabling it will get us running and we'll have to look into the root cause further at some point - thanks!

PR #547, opened 2025-11-19
Sentiment styles

deansallinen:


deansallinen:
I'm happy with how this looks and feels if you want to review

aaroncox:
One thing I just noticed about this is that the optimistic state is being applied before the transaction is successful. If you cancel or it errors, it still changes the state. This was working before I think :thinking_face:

deansallinen:
It _should_ be applying the change before the transaction succeeds - that's the optimism in the optimistic UI.

Because the `derived`  rune is being manually set before the transaction is applied, if the actual value received is different it should revert. How are you testing cancelling/erroring?

aaroncox:
The optimistic approach is AFTER the transaction is broadcast, not when the button is clicked. It's optimistic because its anticipating the transaction to successfully be included in a block on the network, and gives immediate feedback after the Anchor process is completing.

You can reproduce this by triggering the action in Anchor, then cancelling the request in Anchor. This causes the UI to update, and the transaction to never be broacast.

deansallinen:
I see. We're thinking of different types of optimistic UI. I'm going for immediate feedback on user-interaction but, regardless, this is a case that needs to be handled so I'll work on that.

aaroncox:
Going to merge so we have a shared branch and less complications. I know the optimistic UI stuff wasn't exactly what I was thinking and we can always address that in another PR.

PR #551, opened 2025-12-05
governance styling

deansallinen:
- **Fixing pageheader badge styles**
- **Fixing styling on governance page**


aaroncox:
I reviewed the bit that was here, and added a mobile fix, then pushed live (someone was reporting issues). Should be safe to continue on this branch if you had more :+1:

PR #552, opened 2025-12-09
Governance styling

deansallinen:


aaroncox:
These checkboxes are awesome :grin:

aaroncox (approved):
Looks great. I used it on the preview Jungle build with the web authenticator and its got a very nice UX. Also checked out mobile and it looks way better there.

deansallinen:
They're the ones I made for the wharf web-renderer which I'm not sure we ever used :sweat_smile:

aaroncox:
Well we're gonna make good use of em here then! :joy:

PR #553, opened 2025-12-10
Better search errors

deansallinen:
Keeps the application layout when searching and clicking through to a resource (e.g. account, block, contract, etc.) that doesn't exist or cannot be found. Makes it less jarring and a better UX for recovering from typos, etc.

### DIFF
--- a/src/routes/[[locale]]/[network]/(explorer)/+layout.svelte
+++ b/src/routes/[[locale]]/[network]/(explorer)/+layout.svelte
@@ -30,18 +30,22 @@
 
 		return baseBadges;
 	});
+
+	console.log(page.data);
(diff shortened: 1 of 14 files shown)


src/routes/[[locale]]/[network]/(explorer)/+layout.svelte
```diff
@@ -30,18 +30,22 @@
 
 		return baseBadges;
 	});
+
+	console.log(page.data);
```
Comment (aaroncox) [inline src/routes/[[locale]]/[network]/(explorer)/+layout.svelte]: Noticed this one stray console log, but otherwise is this good to go?

deansallinen:
Yeah, should be good to go with that one log line removed

PR #556, opened 2025-12-17
Producers bugfix

deansallinen:


deansallinen:
@aaroncox Does this approach make sense? <https://github.com/greymass/unicove/pull/556/changes#diff-5113c39ff3abd80562c58f0f1a0676b8d071273d340a58b740681f80b2d01c52R62-R65>


Is there a better way of doing this? I was testing on WAX and couldn't load the producers page because of an int64 overflow with the votes.

aaroncox:
I don't think the math works. It's showing only what, ~90m tokens voting? I don't think that's accurate :thinking_face:

aaroncox:
Yeah the problem looks like the decay parameter, it should be different :thinking_face:

aaroncox:
I think this works now:

https://github.com/greymass/unicove/pull/557

Changes the vote decay for WAX (from 52 -> 13) and then just ignores precision to avoid huge numbers.

deansallinen:
Yeah the precision on WAX is crazy. The top BP has `998,269.65118598`

deansallinen:
Hm. Your branch still gives console error: `Uncaught Error: Number 23626948222636083532 overflows int64`

deansallinen:
Never mind, I had to re-make the config. Works now after your branch fix. [](edited-with-Axolo)

